PR #5, opened 2020-01-20
fix nodecon

cgzones:


configure.ac
```diff
@@ -76,5 +76,8 @@ AX_VALGRIND_CHECK
 
 AC_CHECK_HEADER([uthash.h], [], [AC_MSG_ERROR([Unable to find uthash header])])
 
+AM_CFLAGS="-Wall -Wextra -Wmissing-format-attribute -Wmissing-noreturn -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings"
```
dburgener: It looks to me like these aren't actually being used in the build because they are being overridden by the AM_CFLAGS settings in Makefile.am

src/parse.y
```diff
@@ -750,10 +750,12 @@ void yyerror(const char* s) {
 
 	struct check_data data;
 	data.mod_name = get_current_module_name();
-	data.filename = basename(parsing_filename);
+	char *copy = parsing_filename;
```
dburgener: I think you may have intended to make a copy with strdup here?

dburgener:
Are you done pushing to this and ready for me to merge it?  It looks good on my end, although I'm still seeing some warnings in the unit tests, so I wasn't sure if you were done yet or not.  I'm happy to fix those up, but don't want to duplicate work if you're already doing it.

cgzones:
I resolved the last testsuite warnings.
Is the format ok, or do you prefer squashing some similar commits or some other adjustments?

dburgener:
I'm okay with the large number of commits.  If you want to squash that's fine, but I like having a git history that reflects the actual work done.  In the future, I'd find it helpful to break PRs up a little bit more (ie the nodecon fix and the warnings in separate PRs), but for this one, I don't have a problem for merging it as is.  (Once I look over these last two commits of course).

Thanks for all your work on this!

PR #30, opened 2020-02-13
Context flag

dburgener:
@stevedlawrence You had been requesting this feature.  Any thoughts on its implementation?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,12 @@
 ### Added
 - -S flag to print a summary of issue found following an analysis
 - Check S-003 for unneeded semicolons
+- --context flag to specify additional files to parse but not scan.
+  This is primarily helpful if you want to only scan your modified files
+  in a full source repository.  (If you are planning on loading your local
+  modules into your already installed policy and have local development
+  headers in a standard location, selint should find them by default and you
+  don't need to use this option)
 
 ### Fixed
 - Man page generation in distribution tarballs now works after make clean
--- a/README
+++ b/README
@@ -42,6 +42,11 @@ OPTIONS
 		Override default config with config specified on command line.  See
 		CONFIGURATION section for config file syntax.
 
+	--context=CONTEXT_PATH
+		Also parse any .te or .if files found in CONTEXT_PATH and load symbols
+		associated with them for use when checking the policy files to be analyzed.
+		No checks are run on these files.
+
 	-d CHECKID, --disable=CHECKID
 		Disable check with the given ID.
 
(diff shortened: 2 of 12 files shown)


stevedlawrence:
Testing it out, it seems to work, but I'm also seeing some odd behavior.

An example where it's working as I would expect, without context I get the error:

> (E): Nonexistent type (unlabeled_t) listed in fc_entry (E-005

And with context I get

> (S): Type unlabeled_t is declared in module kernel, but used in file context here. (S-002)

However, both with and without context, I see this error for everyone one of my fc contexts:

> (E): Nonexistent user (system_u) listed in fc_entry (E-003)
> (E): Nonexistent role (object_r) listed in fc_entry (E-004)

So even with context, it doesn't seem to know about ``system_u`` and ``object_r``.

Another thing that is odd, is I get very different messages for out of order lines when I provide context. I wouldn't think context would change how it determines order. I'll see if I can create a minimal example that shows what I'm seeing.

stevedlawrence:
Found a way to reproduce using https://github.com/TresysTechnology/clip

Clone that repo, run ``git submodule init && git submodule update``. Then run the following:
```
cd packages/clip-selinux-policy/
diff -Naur \
  <(selint -r clip-selinux-policy) \
  <(selint -r --context refpolicy clip-selinux-policy)
```
There's a handful of different line out of order warnings. Warnings about system layer ordering I guess could make sense since --context might teach selint which layer an interface is in, but others I don't understand why there would be a difference.

dburgener:
I think these are all expected behavior.

The E-003 and E-004 errors are a result of not specifying a config file (either having one in SYSCONFDIR/selint.conf or specified on the comand line with -c).  You're the second person to complain about that, so maybe I need to find some way to make things more clear.  The idea is to make the existence of those configurable rather than just hardcoding them.  Maybe a warning about a lack of a config file would be a helpful addition.

I didn't go through everything in the clip example, but I think those ordering changes are all expected (and for the better).  I see two changes:

- Interface layers, as you notice
- Correct identification of transform interfaces.  For example, without --context it reports an issue on oscap.te:11, but --context makes it go away.  No issue is correct there, because application_domain is a transform interface that is grouped with definitions.  I'm getting a little weirdness in corenet stuff, because corenetwork.if hasn't been created yet, but that's to be expected (it doesn't see corenet interfaces as real interfaces, and decides they must be macros).

I didn't go line by line through everything in clip, but that seems to be the majority of what I see in there.  Do you see something I'm missing?

stevedlawrence:
All makes sense.

Perhaps just having a reasonable built-in config if one isn't provided that assume system_u/object_r  would be helpful. And maybe even hardcode object_r in the check since that's currently hardcoded in the kernel.

+1 :+1:

PR #33, opened 2020-02-15
Add option to exit non-zero on findings

cgzones:


src/main.c
```diff
@@ -457,5 +465,9 @@ int main(int argc, char **argv)
 	free_file_list(context_te_files);
 	free_file_list(context_if_files);
 
+	if (fail_on_finding && found_issue && exit_code == EX_OK) {
+		return EX_DATAERR;
+	}
```
stevedlawrence: Instead of using a ``--fail`` option to change the exit code, what are your thoughts on simplifying this so that we always return EX_DATAERR when there is a finding? I can't think of a case where a zero exit status would be preferred if a problem was found.

We could even do something like pylint [[1]](https://docs.pylint.org/en/1.6.0/run.html#exit-codes) wich returns an exit code that is a combination bit flags of the types of findings, but that extra complication probably isn't be necessary>

dburgener: My inclination is to say that on a normal run with issues found, it successfully ran and reported issues, which makes sense as a zero exit status to me, because the program ran correctly and reported issues as desired.

The ability to have non-zero status as an option makes tons of sense for integration with CI pipelines in particular, but to me it seems like reporting issues is success.

That said, "pylint does it" does seem like a good argument against what I'm saying.  Any idea what other linters do on this topic?

stevedlawrence: Actually, it looks like pylint add an --exit-zero option [[1]](https://github.com/PyCQA/pylint/blob/1.9/ChangeLog#L59), which behaves like selint does now. So someone clearly wanted the current selint behavior. 

And I think golint has the selint behavior [[2]](https://github.com/golang/lint/blob/master/golint/golint.go#L86).

Not sure of others, but it seems not unreasonable to default to zero. And your argument is reasonable. I retract my concern.

cgzones: Also in the C world `cppcheck` and `scan-build` return 0 by default.

I dislike the idea of returning a combination of bit-flags: grepping the output should me more convenient than extracting the desired bit out of the exit value

PR #34, opened 2020-02-15
Fail on error issues when parsing refpolicy in CI

dburgener:


cgzones:
Why was this not merged?

dburgener:
I was concerned that this would become brittle if refpolicy changes upstream caused selint to return error issues for reasons unrelated to the active PR.  There does seem to be some utility in it, so I'm torn.  I take it from the comment, you would prefer to have it in?

cgzones:
I was just curious.
As the main reason (to detect refpolicy parse failures) would cause a ci-run to fail, all is well.

PR #42, opened 2020-02-28
improve ordering

cgzones:
* ignore order of auditallow rules
 * ignore order of interfaces which are of type filetrans and transform
 * order calls to interfaces from own module as section own
 * treat foo, foo_t and foo_r as same raw section
 * accept order of sections bar_t followed by foo_t without any
   interfaces in between
 * print type of issued line
 * add _mountpoint suffix to transform interfaces
 * mark interfaces with associate av-rule always as transform interface
   independent from its name
 * mark interfaces with filetrans_add_pattern or _filetrans calls as
   filtrans interface
 * treat calls to interfaces from mcs and mls module as transform
   interface

### DIFF
--- a/src/ordering.c
+++ b/src/ordering.c
@@ -26,7 +26,7 @@ int is_optional(const struct policy_node *node);
 int is_tunable(const struct policy_node *node);
 int is_in_ifdef(const struct policy_node *node);
 
-struct ordering_metadata *prepare_ordering_metadata(const struct policy_node *head)
+struct ordering_metadata *prepare_ordering_metadata(const struct check_data *data, const struct policy_node *head)
 {
 	const struct policy_node *cur = head->next; // head is file.  Order the contents
 	size_t count = 0;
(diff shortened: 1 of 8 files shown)


dburgener:
Thanks for this!  Very excited to see this because of how much work this section needs.  Just letting you know it may take me a few days before I get a chance to review it.

src/ordering.c
```diff
@@ -450,6 +488,28 @@ enum local_subsection get_local_subsection(const struct policy_node *node)
 	}
 }
 
+/*
+ * Treat the following as the same section:
+ *       foo_t and foo_t
```
dburgener: Should this be foo and foo_r?

src/ordering.c
```diff
@@ -450,6 +488,28 @@ enum local_subsection get_local_subsection(const struct policy_node *node)
...
+ *       foo_t and foo_r
+ *       foo   and foo_t
+ */
+static int is_same_section(const char *first_section_name, const char *second_section_name)
+{
+	size_t first_length = strlen(first_section_name);
+	size_t second_length = strlen(second_section_name);
+
+
+	if (first_length >= 3 && (first_section_name[first_length-1] == 't' || first_section_name[first_length-1] == 'r') && first_section_name[first_length-2] == '_') {
```
dburgener: Can you please break this into multiple lines so that the boolean logic is clearer for readers?  Same with "second_length" equivalent.

dburgener:
@cgzones I pushed up a change to do the current module name in a way that I think is cleaner.  Take a look and let me know what you think.

cgzones:
looks good

PR #50, opened 2020-03-12
Add check for interface calling template

cgzones:


src/if_checks.c
```diff
@@ -237,3 +237,40 @@ struct check_result *check_type_required_but_not_used_in_if(__attribute__((unuse
...
+		parent = parent->parent;
+	}
+	
+	if (!parent) {
+		return NULL;
+	}
+	
+	const struct if_call_data *id = node->data.ic_data;
+	
+	if (parent->flavor != NODE_INTERFACE_DEF && parent->flavor != NODE_TEMP_DEF) {
```
dburgener: This case to me feels more like an internal error.

The most common case seems to be if we are in a block not listed above, which is likely an internal error because we should have listed the block above.

Why not just go up the parents until you find NODE_INTERFACE_DEF or NODE_TEMP_DEF?  If you don't, return NULL.  If you do, proceed.  That gets this check out of the business of listing all the blocks above on line 248.

src/if_checks.c
```diff
@@ -237,3 +237,40 @@ struct check_result *check_type_required_but_not_used_in_if(__attribute__((unuse
...
+		return make_check_result('S',
+					 S_ID_IF_CALLS_TEMPL,
+					 "parent node of %s has unsupported flavor %d",
+					 id->name,
+					 parent->flavor);
+	}
+	
+	if (look_up_in_template_map(id->name) &&
+	    !look_up_in_template_map(parent->data.str)) {
```
dburgener: We could just check the flavor for NODE_TEMP_DEF rather than doing the hash lookup.

src/if_checks.h
```diff
@@ -57,4 +57,16 @@ struct check_result *check_type_required_but_not_used_in_if(const struct
                                                             const struct
                                                             policy_node *node);
 
+/*********************************************
```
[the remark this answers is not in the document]
cgzones: re-ordered

src/if_checks.c
cgzones: re-worked

src/if_checks.c
cgzones: updated to `parent->flavor == NODE_INTERFACE_DEF && look_up_in_template_map(ifd->name)`

Issue #19, opened 2020-02-05
Add check for interface calls using spaces instead of commas


dburgener:
For example:

`foo(bar baz)`

If foo expands to something like:

`allow $1 $2:file perms;`

Then "bar baz" is substituted as $1, and $2 is the empty string and everything works fine.  But obviously strange errors could occur if later the interface is modified to separate $1 and $2.

dburgener:
Added in 9fc3c507b77d3bf4e994670d5549cf4b1aee8718

PR #58, opened 2020-03-16
Add checks for interface arguments

cgzones:


dburgener:
On W-007, I glanced through the output scanning against upstream refpolicy and all the examples of issues found I checked were the false positive on mls arguments.  I'm hesitant to include a check that is so false-positive happy.  Any thoughts on ways to eliminate that?

New functional test files need to be included in tests/Makefile.am FUNCTIONAL_TEST_FILES list or else functional tests won't work out of a distribution tarball.  Looks like there were already some missing before this PR.  I'll get those added in and add a pipeline check for that so that case gets caught earlier.

cgzones:
> On W-007, I glanced through the output scanning against upstream refpolicy and all the examples of issues found I checked were the false positive on mls arguments. I'm hesitant to include a check that is so false-positive happy. Any thoughts on ways to eliminate that?

I am not sure whether these are actual false-positives, do we not want to get warned about such arguments? (see #19)

> New functional test files need to be included in tests/Makefile.am FUNCTIONAL_TEST_FILES list or else functional tests won't work out of a distribution tarball. Looks like there were already some missing before this PR. I'll get those added in and add a pipeline check for that so that case gets caught earlier.

Oops. A pipeline check would be nice.

dburgener:
This feels different to me than the sort of issue I was thinking of in #19.  In the example there, the developer's intent seems likely to be to substitute foo as $1 and bar as $2, when in fact they're substituting "foo bar" as $1 and "" as $2.  In the mls case, the developers intent is to substitute "mls_systemlow - mls_systemhigh" in its entirety as the argument, which is in fact what they're doing.

It seems to me like in this case, the argument for quoting the whole string or whatever the proposed alternative is comes down to making things easier for selint, which doesn't seem like a reasonable demand to make on policy developers.  I could definitely be wrong about that though, but my impression is that the community doesn't see anything wrong with usages like `foo(s0 - s1)`.

An ideal solution might be to compare to the expected number of arguments, but that requires parsing the comments above the interface, which is something I'd like to do eventually, but seems decidedly non-trivial.  A more hacky approach might be to not report this warning if the argument includes a '-' character.  That would also omit "domain -unconfined_t" and the like, which seems desirable to me.

I'm open to hearing arguments on the other side though.

dburgener:
I'm okay with this.  Go ahead and do the squashing and then I'll merge.

PR #67, opened 2020-03-20
Extend name clash check to all declarations

cgzones:


dburgener:
@cgzones You've got conflict on this one.

PR #68, opened 2020-03-20
Add check for usage of bare module statement

cgzones:


src/tree.h
```diff
@@ -172,6 +172,7 @@ struct policy_node {
 	union node_data data;
 	char *exceptions;
 	unsigned int lineno;
+	unsigned char flagged;
```
dburgener: I like making it generic if it's going in the node structure, but I'm worried about whether it will actually be useful beyond this one case, since there's no way of knowing what flagged it.  I'm wondering if a string_list where you added the check id of what it's being flagged for would have more potential for being useful in the future?

The way this sort of thing is handled for other checks is by adding a struct to node_data for that particular node flavor, and then adding a granular flag like "bare_module" there.

cgzones: Added example commit

README
```diff
@@ -155,6 +155,7 @@ CHECK IDS
 	E-004: Nonexistent role listed in fc file
 	E-005: Nonexistent type listed in fc file
 	E-006: Attribute and interface with same name
+	E-007: Module name does not match file name
```
dburgener: I feel like this should be a warning instead of an error.

src/tree.h
dburgener: Yeah, I like this new version better.

README
cgzones: I think `checkpolicy(8)` treats it as error, see https://github.com/SELinuxProject/selinux/blob/9d9a3307de550a4ba31ba41a5e71f3ac27423dbb/checkpolicy/checkmodule.c#L303

dburgener: Interesting.  That's actually good to know.  Last time I encountered this situation must have been before 2016.

In refpolicy builds with modules in base though, the module name will just be discarded, so this check will never be reached.  That feels similar to most of the existing warnings that can cause build time errors in some scenarios but not others.

cgzones: changed to severity warning

cgzones:
Squashed

PR #69, opened 2020-03-20
Implement roleattribute and attribute_role statements

cgzones:


src/tree.h
```diff
@@ -159,7 +167,7 @@ union node_data {
 	struct if_call_data *ic_data;
 	struct declaration_data *d_data;
 	struct fc_entry *fc_data;
-	struct type_attribute_data *ta_data;
+	struct attribute_data *ta_data;
```
dburgener: This probably shouldn't be named "ta_data" anymore if the struct isn't named type_attribute_data anymore.

src/ordering.c
```diff
@@ -541,6 +542,7 @@ enum order_difference_reason compare_nodes_refpolicy_generic(struct ordering_met
 			if (first->data.d_data->flavor != second->data.d_data->flavor) {
 				CHECK_FLAVOR_ORDERING(d_data, DECL_BOOL, ORDER_DECLARATION_SUBSECTION);
 				CHECK_FLAVOR_ORDERING(d_data, DECL_ATTRIBUTE, ORDER_DECLARATION_SUBSECTION);
+				CHECK_FLAVOR_ORDERING(d_data, DECL_ATTRIBUTE_ROLE, ORDER_DECLARATION_SUBSECTION);
```
dburgener: The style guide doesn't specify the order of attribute_role declarations in the declarations section.  I think this can just be omitted.

src/maps.c
```diff
@@ -20,6 +20,7 @@ struct hash_elem *type_map = NULL;
 struct hash_elem *role_map = NULL;
 struct hash_elem *user_map = NULL;
 struct hash_elem *attr_map = NULL;
+struct hash_elem *roleattr_map = NULL;
```
dburgener: Is there a reason this is named "roleattr" when it's a map of "attribute_role"s rather than roleattributes?

src/tree.h
cgzones: changed to `at_data`

src/ordering.c
cgzones: dropped

src/maps.c
cgzones: changed

PR #71, opened 2020-03-23
Fixes

cgzones:


src/parse_fc.c
```diff
@@ -36,6 +36,10 @@ struct fc_entry *parse_fc_line(char *line)
 
 	char *pos = strtok(line, whitespace);
 
+	if (pos == NULL) {
```
dburgener: What's the motivation on this?  I'd been thinking it actually would make more sense to go the opposite direction and get te and if files to recover on errors where possible.  The common case I'm envisioning for this is when a user has typed something like "gen_conetxt" or something, which is definitely an error case, but seems like a good scenario to report on and then continue parsing.

cgzones: I think one must not pass `NULL` to `strdup()`.
Do you prefer the diff:
```
- out->path = strdup(pos);
+ out->path = pos ? strdup(pos) : NULL;
```

dburgener: My bad.  I was confused about where in the code this was.  Makes perfect sense now, thanks!

PR #73, opened 2020-03-24
Separate role-types statement from role-declaration

cgzones:
Add a separate new node type for role-types statements, e.g.
    `role bar_r type foo_t;`

### DIFF
--- a/src/ordering.c
+++ b/src/ordering.c
@@ -193,7 +193,8 @@ const char *get_section(const struct policy_node *node)
 	case NODE_RT_RULE:
 		return "_non_ordered";
 	case NODE_ROLE_ALLOW:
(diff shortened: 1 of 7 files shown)


src/ordering.c
```diff
@@ -193,6 +193,7 @@ const char *get_section(const struct policy_node *node)
 	case NODE_RT_RULE:
 		return "_non_ordered";
 	case NODE_ROLE_ALLOW:
+	case NODE_ROLE_TYPES:
 		// This is not in the style guide. I normally see it grouped
```
dburgener: I think comment is true for both.  Can you change it to a "these are" and "them" now that it refers to two items?

cgzones: updated comment

PR #75, opened 2020-03-24
Add check for unquoted gen_require blocks

cgzones:


tests/functional/end-to-end.bats
```diff
@@ -267,7 +271,7 @@ test_one_check() {
 
 	run ${SELINT_PATH} -c configs/default.conf policies/misc/disable_require_start.*
 	[ "$status" -eq 0 ]
-	[ "$output" == "" ]
+	[ "$output" == "disable_require_start.te:8: (W): Gen require block unquoted (W-010)" ]
```
dburgener: Can you just add a "-d W-010" to the selint call instead?

cgzones: yup

README
```diff
@@ -151,6 +151,7 @@ CHECK IDS
 	W-007: Unquoted space in argument of interface call
 	W-008: Allow rule with complement or wildcard permission
 	W-009: Module name does not match file name
+	W-010: Unquoted gen_require block
```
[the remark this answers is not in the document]
cgzones: maybe some macro expansion issues, but style severity is fine with me

README
```diff
@@ -141,6 +141,7 @@ CHECK IDS
 	S-004: Template call from an interface
 	S-005: Declaration in interface
 	S-006: Bare module statement
+	S-007: Call to gen_context omits mls component
```
dburgener: Good catch, thanks

PR #80, opened 2020-03-25
RFC: Add check for multiple classes in allow rule

cgzones:
Findings in refpolicy:

```
init.te:        246: (C): Allow rule with more than one class (C-006)
    allow init_t init_path_unit_loc_type:{ dir file } { getattr watch };
bootloader.te:  162: (C): Allow rule with more than one class (C-006)
    allow bootloader_t bootloader_tmp_t:{ dir file } { relabelfrom relabelto };
courier.te:     140: (C): Allow rule with more than one class (C-006)
    allow courier_pop_t courier_tcpd_t:{ unix_stream_socket tcp_socket } rw_stream_socket_perms;
postfixpolicyd.te:40: (C): Allow rule with more than one class (C-006)
    allow postfix_policyd_t postfix_policyd_tmp_t:{ file sock_file } manage_file_perms;
xserver.te:     805: (C): Allow rule with more than one class (C-006)
    allow xserver_t xevent_type:{ x_event x_synthetic_event } { send receive };
xserver.te:     950: (C): Allow rule with more than one class (C-006)
    allow x_domain self:{ x_device x_pointer x_keyboard } { getattr setattr use read write getfocus setfocus bell force_cursor freeze grab manage list_property get_property set_property add remove create destroy };
xserver.te:    1015: (C): Allow rule with more than one class (C-006)
    allow x_domain xevent_type:{ x_event x_synthetic_event } { send receive };
xserver.te:    1032: (C): Allow rule with more than one class (C-006)
    allow xserver_unconfined_type xevent_type:{ x_event x_synthetic_event } { send receive };
domain.te:      159: (C): Allow rule with more than one class (C-006)
    allow unconfined_domain_type domain:{ socket_class_set socket key_socket } { create_stream_socket_perms lock relabelto name_bind map sendto recvfrom relabelfrom };
unconfined.if:   23: (C): Allow rule with more than one class (C-006)
    llow $1 self:{ capability cap_userns } { chown dac_override dac_read_search fowner fsetid kill setgid setuid setpcap linux_immutable net_bind_service net_broadcast net_admin net_raw ipc_lock ipc_owner sys_rawio sys_chroot sys_ptrace sys_pacct sys_admin sys_boot sys_nice sys_resource sys_time sys_tty_config mknod lease audit_write audit_control setfcap };
unconfined.if:   24: (C): Allow rule with more than one class (C-006)
    allow $1 self:{ capability2 cap2_userns } { syslog wake_alarm };
xserver.if:    1595: (C): Allow rule with more than one class (C-006)
    allow $1 xserver_t:{ x_device x_pointer x_keyboard } { getattr setattr use read write getfocus setfocus bell force_cursor freeze grab manage list_property get_property set_property add remove create destroy };
```

### DIFF
--- a/README
+++ b/README
@@ -134,6 +134,7 @@ CHECK IDS
 	C-001: Violation of refpolicy te file ordering conventions
 	C-004: Interface does not have documentation comment
 	C-005: Permissions in av rule or class declaration not ordered
+	C-006: Allow rule with multiple classes
 
 	S-001: Require block used instead of interface call
 	S-002: File context file labels with type not declared in module
--- a/src/check_hooks.h
+++ b/src/check_hooks.h
@@ -25,6 +25,7 @@ enum convention_ids {
 	C_ID_TE_ORDER       = 1,
 	C_ID_IF_COMMENT     = 4,
 	C_ID_UNORDERED_PERM = 5,
+	C_ID_MUL_CLS_ALLOW  = 6,
 	C_END
 };
 
--- a/src/runner.c
+++ b/src/runner.c
@@ -141,6 +141,10 @@ struct checks *register_checks(char level,
 			add_check(NODE_DECL, ck, "C-005",
 			          check_unordered_perms);
 		}
+		if (CHECK_ENABLED("C-006")) {
+			add_check(NODE_AV_RULE, ck, "C-006",
+			          check_multi_class_allow_rule);
+		}
 		// FALLTHRU
 	case 'S':
 		if (CHECK_ENABLED("S-001")) {
--- a/src/te_checks.c
+++ b/src/te_checks.c
@@ -114,6 +114,35 @@ struct check_result *check_unordered_perms(__attribute__((unused)) const struct 
 
 }
 
+struct check_result *check_multi_class_allow_rule(__attribute__((unused)) const struct check_data *data,
+                                                  const struct policy_node *node)
+{
+	// ignore non-allow rules
+	if (node->data.av_data->flavor != AV_RULE_ALLOW) {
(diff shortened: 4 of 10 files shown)


dburgener:
I'm okay with the general idea, but I feel like more exceptions are warranted.

For example, in some of these cases the permission set is identical between the two object classes (for example capability and cap_userns).  I don't see anything wrong with the grouping in at least some of those cases.

An exception to the exception might be a situation like lnk_file and sock_file where the pemissions are the same, but the macros in obj_perm_sets.spt are not identical.

Maybe just list capability/cap_userns and capability2/cap2_userns as exceptions?  Maybe some of the IPC mechanisms make sense for that as well?

A second area that I think might make sense for exceptions is userspace object classes.  Those tend to be isolated in one module, and can have somewhat unique usage patterns.  Glancing at a few of the xserver ones, I don't see a problem with the usage there.  They've defined a lot of very similar object classes with identical permission sets, and grouping them that way makes sense.

Lastly, I'm wondering if 2 permissions instead of 1 makes sense for an exception.  I feel like usages like the { getattr watch } and { relabelfrom relabelto } listed above seem likely to be stable and not cause any maintenance problems going forward.  What about making that value configurable and setting the default to 2?

I'm very open to discussion on any of the points above if you think otherwise.

cgzones:
I not sure anymore this is really something one needs.
Maybe improving `S-009` (Permission macro suffix does not match class name) to check for something like `allow source target:{ file fifo_file } read_file_perms;` is enough.

cgzones:
Superseded by #148

Issue #27, opened 2020-02-05
Order output by module


dburgener:
It seems like it would be easier to read and understand that way.

PR #82, opened 2020-03-30
Increase padding for filenames to 22

cgzones:
Refpolicy modules with long names:

postfixpolicyd
networkmanager
setroubleshoot


dburgener:
In the long term, I'd like to move to dynamic padding based on the length of the filenames outputted.  That would require caching issues found and then displaying them all at once.  That caching is also required for #27 I believe.

PR #84, opened 2020-03-30
Improve handling of boolean declarations

cgzones:


dburgener:
Does this change mean that src/ordering.c lines 238 and 239 can be removed?

Also, I don't see anything in any test code validating our gen_bool handling.  Would you be willing to add a test case in check_parsing.c to validate that this is all handled correctly (parsing, declarations stored etc)?

cgzones:
yup, will update

PR #86, opened 2020-03-30
Add colored output

cgzones:


cgzones:
Added README section

src/color.c
```diff
@@ -0,0 +1,99 @@
+/*
+* Copyright 2020 Tresys Technology, LLC
```
dburgener, line 2: I am not a lawyer, and could definitely be wrong about this but I think the copyright of the code is owned by the person who wrote it (or their employer if produced as a work for hire), and is not governed on a whole project basis.  Given that you are the sole author of this file, I believe this is your copyright, not Tresys's.  My understanding is that you have the option but not the responsibility to mark your copyright where applicable.  I don't think it's correct to indicate that this file (or color.h) is copyright Tresys though, since the code was not written as a work for hire by a Tresys employee.  It is my understanding that you can leave it blank or mark your own copyright as you see fit.

As I understand it, each contributor (or their employer) owns the copyright for the portions of the code they authored, regardless of marking.  The Copyright notice is just a helpful indicator in that regard.  FWIW here's the Linux Foundation's opinion on the general matter of copyrights: https://www.linuxfoundation.org/blog/2020/01/copyright-notices-in-open-source-software-projects/

ping @stevedlawrence for your thoughts?

src/color.h
```diff
@@ -0,0 +1,63 @@
...
+const char *color_ok(void);
+
+/*********************************************
+* Mark the following output with the appropiate color for the given severity.
+* severity - The severity to decide the color.
+*********************************************/
+const char *color_severity(char severity);
+
+
+
```
dburgener, line 60: Why so much white space?

src/check_hooks.c
```diff
@@ -22,6 +22,8 @@
 
 #include "check_hooks.h"
```
dburgener, line 24: No newline here please

src/main.c
```diff
@@ -146,9 +152,22 @@ int main(int argc, char **argv)
...
 
+		case COLOR_ID:
+			if (0 == strcmp(optarg, "on")) {
+				color = 2;
+			} else if (0 == strcmp(optarg, "off")) {
+				color = 1;
+			} else if (0 == strcmp(optarg, "auto")) {
+				color = 0;
+			} else {
+				printf("invalid argument '%s' given for option --color\n", optarg);
```
dburgener, line 163: All our other messages begin with a capital letter

src/main.c
```diff
@@ -350,10 +388,15 @@ int main(int argc, char **argv)
 			// TODO: Make modules.conf name configurable
 			modules_conf_path = strdup(file->fts_path);
 		} else {
-			print_if_verbose(
-				"Skipping %s which is not a policy file\n",
-				file->fts_path);
-			if (!recursive_scan) {
+			if (recursive_scan) {
```
dburgener, line 391: The idea is to make the message more highlighted and unconditional in non-recursive mode?  That makes sense to me, but it too me a while to figure out what the intention was here.  Can you comment this a little more thoroughly to explain the rationale behind the recursive/non-recursive difference in output?

I do feel like we should avoid printing the message twice for directories in both modes though.

dburgener:
Thanks for this contribution.  It looks really nice with colored output.

src/color.c
stevedlawrence: This is in line with my understanding.

cgzones: Thanks for the advice.
I changed the copyright note to `Copyright 2020 The SELint Contributors`.

src/color.h
cgzones: removed

src/check_hooks.c
cgzones: removed

src/main.c
cgzones: Capitalized

src/main.c
cgzones: Reformatted this logic; I hope it is more readable now

PR #87, opened 2020-03-30
Use fts_open flag FTS_XDEV

cgzones:
Do we want to cross device boundaries?

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -320,7 +320,7 @@ int main(int argc, char **argv)
 
 	paths[i] = NULL;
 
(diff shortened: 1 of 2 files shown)


dburgener:
What is the benefit to enforcing this?

It's undoubtedly an uncommon scenario that this would even come up in.  Is there a plausible scenario where the user didn't intentionally set this up?  If so, what is the impact?  If the user did set this up themselves, I think they would expect their files to be scanned across device boundaries.

cgzones:
fair enough

PR #88, opened 2020-04-02
Allow selint-disable comment after declaration in require block

cgzones:
Fixes: 3f83758c1825c1e720867e18ab97ee18ef0eadf9 ("Separate grammar for real declarations and require content")

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -469,6 +469,12 @@ require_lines:
 	;
 
 require_line:
+	require_bare
+	|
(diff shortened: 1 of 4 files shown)


dburgener:
Something weird is going on with actions/checkout in the integration pipeline on this PR.  It was originally failing on an intermittent github error, but now on rerun, it's not fetching the pr ref.

I noticed that there is an actions/checkout v2 so I updated to that.  Try rebasing on it and see if that fixes the integration pipeline issue.

PR #93, opened 2020-04-06
Add option to only display summary

cgzones:


README
```diff
@@ -76,8 +76,13 @@ OPTIONS
...
 		will assume that scanned policy files are intended to be loaded into the
 		currently running system policy.
+
 	-S, --summary
-		Display a summary of issues found after running the analysis
+		Display a summary of issues found after running the analysis.
+
+	--summary-only
+		Only display a summary of issues found after running the analysis.
+		Do not show the individual findings.
```
dburgener, line 85: I think we should add "Implies -S" at the end to avoid possible confusion.

src/main.c
```diff
@@ -67,6 +68,8 @@ static void usage(void)
 		"  -s, --source\t\t\tRun in \"source mode\" to scan a policy source repository\n"\
 		"\t\t\t\tthat is designed to compile into a full system policy.\n"\
 		"  -S, --summary\t\t\tDisplay a summary of issues found after running the analysis\n"\
+		"      --summary-only\t\tOnly display a summary of issues found after running the analysis\n"\
```
dburgener, line 71: Can you add the period at the end of this line as well?  It makes the man page output look weird without it.

PR #94, opened 2020-04-06
Add cleanup destructors for parser allocations

cgzones:
Avoid memory leaks on ill-formed inputs.

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -140,6 +140,9 @@
 %type<av_flavor> av_type
 %type<node_flavor> if_keyword
 
+%destructor { free($$); } mls_component mls_level mls_range sl_item
+%destructor { free_string_list($$); } arg args comma_string_list string_list strings
+
 %%
 selinux_file:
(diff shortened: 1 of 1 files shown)


src/parse.y
```diff
@@ -56,6 +56,9 @@
 %destructor { free($$); } QUOTED_STRING
 %destructor { free($$); } VERSION_NO
 
+%destructor { free($$); } mls_component mls_level mls_range sl_item
```
dburgener, line 59: Isn't an sl_item just a string_list with one node?

cgzones: I think its just a string (no `struct string_list`)
https://github.com/TresysTechnology/selint/blob/9221bd945e4fdc0035c063f08778d13b13febcd5/src/parse.y#L353-L363

dburgener: You're right.  I see now why I was confused.  I was getting over a cold on Tuesday morning, so I blame that.

PR #96, opened 2020-04-09
Skip S-002 checks on probably build generated filecontexts

cgzones:
When running SELint on a already build policy structure,
S-002 triggers on the files base.fc, all_mods.fc and $MODULENAME.mod.fc

### DIFF
--- a/selint.conf
+++ b/selint.conf
@@ -50,4 +50,5 @@ ordering_rules = "refpolicy-lax"
 # Currently the following checks are going to be disabled on build generated
 # files:
 #     S-002: base.fc, all_mods.fc, $MODNAME.mod.fc
+# If unset, this defaults to true.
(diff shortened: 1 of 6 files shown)


cgzones:
Feel free to skip 2nd commit. It prints a one-time note that S-002 is not checked on build generated filecontext files.

dburgener:
Can you add a config option to disable this behavior, and point people to it in the one-time note?  Definitely have in enabled by default, but I'm envisioning someone really wanting a module named "all_mods" for some reason wanting to be able to disable this.

cgzones:
Better names than `ignore_build_fps` are welcome

dburgener:
How about "skip_checking_generated_fcs"?

I'd also vote for having it commented out by default.  That might help communicate to ordinary users that this isn't something they need to worry about.

PR #99, opened 2020-04-11
Declare global variable in header file

cgzones:


dburgener:
Why expose this variable to more files?  I don't see why anything other than main and util would ever need to access this variable, so I don't see the benefit to putting it in the header.

cgzones:
fair enough

PR #105, opened 2020-04-11
Ignore intentional const discards also on clang

cgzones:


cgzones:
Added a clang run to Github actions

dburgener:
Github actions has support for something called a build matrix (https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#configuring-a-build-matrix) where you can have the same set of steps run on multiple compilers without copy/pasting code.

@stevedlawrence has some code for that already that he can make a PR for.  How would you feel about scaling this back to just the other improvements and then Steve can add the build matrix portion to get clang compilation in the pipeline?

cgzones:
Dropped all GitHub action related changes.

PR #113, opened 2020-04-14
Skip hidden directories by default and option to scan them

cgzones:


src/main.c
```diff
@@ -408,6 +418,11 @@ int main(int argc, char **argv)
 			if (!recursive_scan) {
 				fts_set(ftsp, file, FTS_SKIP);
 			}
+
+			if (!scan_hidden_dirs && file->fts_info == FTS_D && file->fts_name[0] == '.' && file->fts_name[1] != '.' && file->fts_name[1] != '\0') {
```
dburgener, line 422: Can you split this into multiple lines so it's more readable?

cgzones: done

PR #121, opened 2020-04-17
Check for conversion compiler warnings

cgzones:
Exclude them from Werror on CI runs though, cause the ubuntu bison/flex versions do not generate warning free code


dburgener:
Looks fine.  Want to run it on fedora before merging, but I'm on my Ubuntu box right now.

PR #124, opened 2020-04-20
Pretty parser

cgzones:
* Use Lex option `noyywrap` instead of implementing `yywrap()`
* Drop explicit lexer rule for unknown `%` character
* Add Lex option `nodefault`
* Fix Lex catch-all rule
* Use Lex option `yylineno` instead of manual handling
* Enable more verbose syntax error messages

### DIFF
--- a/src/lex.l
+++ b/src/lex.l
@@ -108,4 +108,10 @@ userdebug_or_eng { return USERDEBUG_OR_ENG; }
 (\#.*)?\n { yylineno++; }
 dnl(.*)?\n { yylineno++; }
 [ \t] ; /* normally skip whitespace */
-. { char *str = malloc(strlen(yytext) + 1); sprintf(str, "Invalid character: %s", yytext); yyerror(str); free(str); }
+. {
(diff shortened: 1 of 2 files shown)


src/lex.l
```diff
@@ -107,5 +107,4 @@ userdebug_or_eng { return USERDEBUG_OR_ENG; }
 (\#.*)?\n { yylineno++; }
 dnl(.*)?\n { yylineno++; }
 [ \t] ; /* normally skip whitespace */
-\% { yyerror("Invalid character: %%"); /* will be run through printf again */ }
```
dburgener, line 109: This was added because yyerror expects format strings, so "Invalid character: %", which is what the generic case would return would cause an error on printf in display_check_result.

dburgener:
Nevermind, my bad. On further testing, it looks like the generic case handles the % situation now.  Let me do some more investigation to figure out when that got fixed, then I'll merge the whole series.

dburgener:
Ah, looks like be44a0c4 fixed the underlying bug, so now the generic case works fine for percent.

Thanks for the fixes!

PR #132, opened 2020-05-02
Parse fc macros

dburgener:


selint.conf
```diff
@@ -37,6 +37,10 @@ enable_source = { E-003, E-004 }
 assume_users = { system_u }
 assume_roles = { object_r }
 
+# If you have defined any custom macros for use in fc files, list them here,
+# otherwise, SELinux will report E-002 on occurances of them
```
cgzones, line 41: s/SELinux/SELint/

src/check_hooks.c
```diff
@@ -55,7 +55,7 @@ enum selint_error add_check(enum node_flavor check_flavor, struct checks *ck,
 	return SELINT_SUCCESS;
 }
 
-enum selint_error call_checks(struct checks *ck,
+enum selint_error call_checks(const struct checks *ck,
```
cgzones, line 58: Do we want this to be marked `const`?
We pass the member `check_nodes` to `call_checks_for_node_type()`, which modifies the member

src/selint_config.c
```diff
@@ -76,6 +76,7 @@ enum selint_error parse_config(const char *config_filename,
                                char *severity,
                                struct string_list **config_disabled_checks,
                                struct string_list **config_enabled_checks,
+			       struct string_list **custom_fc_macros,
```
cgzones, line 79: tabs used instead of spaces like the other function arguments

dburgener:
Thanks for the review!  I'll get these cleaned up ASAP.

PR #134, opened 2020-05-04
Fix various small issues with fc macros implementation

dburgener:


dburgener:
@cgzones Did I miss anything?

cgzones:
lgtm

PR #140, opened 2020-05-14
Add unit test for declarations in template

cgzones:
I don't know why valgrind is not issuing the ast leak in test_nested_template_declarations currently.

### DIFF
--- a/src/parse_functions.c
+++ b/src/parse_functions.c
@@ -92,7 +92,7 @@ enum selint_error insert_declaration(struct policy_node **cur,
 		// In a require block, the objects aren't being declared
 		// Otherwise, we need to insert them into the appropriate map
 
-		char *temp_name = get_name_if_in_template(*cur);
+		const char *temp_name = get_name_if_in_template(*cur);
 
 		if (temp_name) {
 			// We are inside a template, so we need to save declarations in the template map
@@ -152,7 +152,7 @@ enum selint_error insert_aliases(struct policy_node **cur,
 	struct string_list *alias = aliases;
 
 	while (alias) {
-		char *temp_name = get_name_if_in_template(*cur);
+		const char *temp_name = get_name_if_in_template(*cur);
 		if (temp_name) {
 			insert_decl_into_template_map(temp_name, flavor,
 			                              alias->string);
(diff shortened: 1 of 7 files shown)


dburgener:
> I don't know why valgrind is not issuing the ast leak in test_nested_template_declarations currently.

That's concerning.  I'll see what I can figure out.

dburgener:
Okay, I think it's because that memory is still reachable at the end of the function, and by default valgrind only shows definitely lost memory and not still reachable.  I had thought that I had --show-reachable enabled at least in the functional tests, but it looks like not.

I'll plan on definitely enabling that on functional tests and see if there's a good way to get them in unit tests as well (need to reread all of the automake docs about that sort of thing)

cgzones:
yeah, seems like we need to add `VALGRIND_memcheck_FLAGS=--leak-check=full --show-reachable=yes --show-leak-kinds=all --errors-for-leak-kinds=all` to tests/Makefile.am

dburgener:
That's roughly where I got as well.  I had thought that any of the last three flags should be sufficient by itself though?

However, I put a leak that I'm pretty sure is non-reachable in a unit test and I still can't get a failure.  I think that possibly the CK_FORK=no environmental variable may be getting lost when the top level make calls make -C tests.

dburgener:
Scratch that theory, I think.  I had a unit test check the environmental variable, and it's definitely still set correctly under make check-valgrind.

Right now I've got this hunk in check_maps.c for testing purposes:

```
@@ -142,6 +144,10 @@ END_TEST
 
 START_TEST (test_insert_decl_into_template_map) {
 
+       char *tmp = malloc(1);
+       *tmp = '\0';
+       tmp = NULL;
+
        insert_decl_into_template_map("user_domain", DECL_TYPE, "$1_t");
        insert_decl_into_template_map("user_domain", DECL_TYPE, "$1_exec_t");
        insert_decl_into_template_map("user_domain", DECL_ROLE, "$1_r");
```

And I can't get make check-valgrind to complain about it at all.

dburgener:
Okay, I tried your full set of flags and now it's working.  I guess I was wrong about them being optional.

We do have several additional leaks that this turns up.  I'll get them fixed and then push the additional valgrind options along with the fixes.

PR #145, opened 2020-05-19
Add configure option --enable-werror

cgzones:
Treat compiler warnings as errors by passing -Werror
Avoids passing -Werror in CFLAGS environment variable,
which might break configure test compilations

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,7 +61,7 @@ jobs:
       run: ./autogen.sh
     - name: configure
       # old flex/bison versions might not generate conversion warning free code
-      run: ./configure CFLAGS="-Werror -Wno-error=conversion -Wno-error=sign-conversion"
+      run: ./configure --enable-werror CFLAGS="-Wno-error=conversion -Wno-error=sign-conversion"
     - name: make
       run: make
     - name: make check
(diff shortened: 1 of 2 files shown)


dburgener:
What's the actual issue with the existing approach?  What sort of test configuration is it breaking?

cgzones:
The configure step (`./configure ....`) compiles several test binaries to check for miscellaneous system dependencies. If the `-Werror` and maybe additional warning flags are passed via `CFLAGS`, these test compilations might undesirable fail.
So for packaging I prefer to call `./configure --enable-werror` instead of e.g. `export DEB_CFLAGS_MAINT_APPEND = -Werror` (Debian)

p.s.: the release tarball contains the pre-build files `lex.c` and `parse.c`. Is this by design?

dburgener:
Okay, that seems reasonable.  It seems like if we're moving -Werror into configure.ac, we probably should move the two Wno-error flags as well?

Yes, the lex.c and parse.c inclusion is by design.  I believe that's default autotools behavior.  The purpose is to avoid requiring yacc/bison dependencies on those downloading the release tarball.  The warnings are definitely annoying though.  Maybe the best solution might be to upgrade bison in the pipeline like we do with bats and check?

PR #147, opened 2020-05-26
[WIP] Add check for invalid statements in blocks

cgzones:
Compatibility assumed from https://selinuxproject.org/page/PolicyLanguage


cgzones:
Maybe this is not sensible to add after all:
Ultimately the SELinux compiler will fail to build such policies and we only detect simple nested violations, e.g.
```
if(bool_name) {
  allow role1 role2;
}
```
but not
```
interface(`interface_name',`
  allow $1 $2;
')
if(bool_name) {
  interface_name(role1, role2)
}
```

PR #154, opened 2020-07-30
Parse access vectors and check for unknown classes and permission

cgzones:


src/te_checks.c
```diff
@@ -611,39 +615,108 @@ struct check_result *check_declaration_interface_nameclash(__attribute__((unused
...
 {
-	static unsigned int permmacros_count = (unsigned int)-1;
-
-	if (permmacros_count == (unsigned int)-1) {
-		permmacros_count = permmacros_map_count();
+	static bool notified = false;
+	// ignore if no permission or permission macro have been parsed
+	if (permmacros_map_count() == 0) {
+		if (!notified) {
+			printf("%sNote%s: Check E-007 is not performed because no permission macro has been parsed.\n",
```
dburgener, line 627: Any particular reason to have this logic in the check rather than at enablement time?

I'm wondering about a general model where if a check needs particular resources to be available, we disable it while registering checks and notify at that point, instead of the first time it encounters it.

tests/functional/policies/check_triggers/e07.pass.te
```diff
@@ -1,3 +1,10 @@
-policy_module(w12, 1.0)
+policy_module(e07_pass, 1.0)
+
+allow source target:file read;
+
+allow source target:file { open read };
 
 allow source target:file read_file_perms;
+
+# TODO:
```
dburgener, line 9: What's the TODO for?

tests/Makefile.am
```diff
@@ -276,7 +277,7 @@ check_maps_SOURCES = check_maps.c ${MAPS_HEADS}
 check_maps_LDADD = @CHECK_LIBS@ $(sort ${MAPS_OBJS})
 
 check_startup_SOURCES = check_startup.c ${STARTUP_HEADS} ${MAPS_HEADS} ${SELINT_ERROR_HEADS}
-check_startup_LDADD = @CHECK_LIBS@ $(sort ${STARTUP_OBJS} ${MAPS_OBJS})
+check_startup_LDADD = @CHECK_LIBS@ $(sort ${STARTUP_OBJS} ${MAPS_OBJS} ${RUNNER_OBJS})
```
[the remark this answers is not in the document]
cgzones: Seems to be a leftover from developing.
Seems to work without `${RUNNER_OBJS}` at all (`${STARTUP_OBJS}` already includes `${PARSE_OBJS}`)

src/te_checks.c
cgzones: I'll try to refactor

tests/functional/policies/check_triggers/e07.pass.te
cgzones: The class `file` does not support the permission `search`.
But since this kind of check is not yet implemented, this statement still passes and is therefore in the `pass.te` file.
I'll improve the comment.

PR #157, opened 2020-08-13
Add support for extended permission rules

cgzones:
Like
    allowxperm basic_t basic_dev_t:chr_file ioctl ~0x8927;

TODO: validate range usage: enforce two surrounding increasing elements

### DIFF
--- a/src/lex.l
+++ b/src/lex.l
@@ -39,9 +39,13 @@ role { return ROLE; }
 types {  return TYPES; }
 attribute_role { return ATTRIBUTE_ROLE; }
 allow { return ALLOW; }
+allowxperm { return ALLOW_XPERM; }
 auditallow { return AUDIT_ALLOW; }
+auditallowxperm { return AUDIT_ALLOW_XPERM; }
 dontaudit { return DONT_AUDIT; }
+dontauditxperm { return DONT_AUDIT_XPERM; }
 neverallow { return NEVER_ALLOW; }
+neverallowxperm { return NEVER_ALLOW_XPERM; }
 type_transition { return TYPE_TRANSITION; }
 type_member { return TYPE_MEMBER; }
 type_change { return TYPE_CHANGE; }
--- a/src/ordering.c
+++ b/src/ordering.c
@@ -179,6 +179,7 @@ const char *get_section(const struct policy_node *node)
 	case NODE_HEADER:
 		return SECTION_NON_ORDERED; // Guaranteed at top by grammar
 	case NODE_AV_RULE:
+	case NODE_XAV_RULE:
 		if (node->data.av_data->flavor == AV_RULE_NEVERALLOW) {
 			// These are somewhat of a unique situation, and the style guide
 			// doesn't mention them explicitly.  Maybe they should just group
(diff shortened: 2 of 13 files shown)


dburgener:
What do you think about this slightly modification?

https://github.com/dburgener/selint/tree/dburgener/x_perms

I think it reduces code complexity and helps ensure that the x_perm version will stay in sync with regular av_rules where applicable.

cgzones:
I think this doesn't work, since in the combined cases `node->data.av_data` is always accessed (and not `node->data.xav_data`).
One could make `char *operation` a part of `struct av_data` and set it to NULL in normal av rules.

dburgener:
But since node->data is a union, av_data and xav_data are just different names for the same memory.  Since my commit reorders the first elements in xav_rule_data to match av_rule_data and C guarantees that structs are stored in order, if we call an xav_rule_data pointer an av_rule_data pointer and access its members they'll end up being the same, as long as the entirety of the av_rule_data struct is at the beginning of the xav_rule_data_struct.

Since in the two cases I changed we only need the av_rule portion of the struct, it should all just work pretending to be av_rule_data.

Running in gdb to demonstrate

 ```
Breakpoint 1, get_names_in_node (node=node@entry=0x5555570fc580) at tree.c:165
165			av_data = node->data.av_data;
(gdb) p node->data.av_data
$1 = (struct av_rule_data *) 0x5555570fa000
(gdb) p node->data.xav_data
$2 = (struct xav_rule_data *) 0x5555570fa000
(gdb) p *(node->data.av_data)
$3 = {flavor = AV_RULE_ALLOW, sources = 0x5555570fc480, targets = 0x5555570fc4a0, object_classes = 0x5555570fc4e0, perms = 0x5555570fc520}
(gdb) p *(node->data.xav_data)
$4 = {flavor = AV_RULE_ALLOW, sources = 0x5555570fc480, targets = 0x5555570fc4a0, object_classes = 0x5555570fc4e0, perms = 0x5555570fc520, operation = 0x51 <error: Cannot access memory at address 0x51>}
```

You see that av_data and xav_data are pointers to the same address, and but appear as different structs when derefenced. In this particular case, the node is a regular allow rule, so the contents of operation is undefined, but the other fields are all identical across both structs.

cgzones:
You're right, forgot about the union stuff.
Added some more verbose testing.

dburgener:
Great.  I'll go ahead and squash and then merge it.  Thanks!

cgzones:
p.s.: the validation of ranged extended permissions is still missing.
E.g.
  `0xffff-0x0000`
or
  `- 0x01`
or
  `0x01 -`
, but this needs quite some code, probably also to derive permission macros.
So for now let's rely on the SELinux compiler for such things.

dburgener:
Yeah, that's fine with me.  In my opinion, we're primarily looking to add value by getting things the compiler can't/doesn't.  We don't need to bite everything off all at once.

I very much want to be in a state where we can parse anything the compiler does (which this PR helps with), but if the compiler fails things that we allow, that's fine as far as I'm concerned.

Not that I wouldn't take a PR for that validation, but it's not high on my priority list.

PR #158, opened 2020-08-20
Allow selint-disable commands more space flexibility.

dburgener:
Specifically allow a space before the check ID, and between check IDs in a list.

Before and After; #selint-disable:S-001,S-002
After: #selint-disable: S-001, S-002


cgzones:
LGTM, maybe add an example to the `SELINT EXCEPTIONS` section in the README?

PR #159, opened 2020-08-25
Add configuration for LGTM.com

cgzones:
Everything besides the build-dependency autoconf-archive gets
automatically detected.

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -1,0 +1,4 @@
+extraction:
+  cpp:
+    prepare:
+      packages: "autoconf-archive"


dburgener:
Should I be seeing a LGTM check in the automated pipeline as a result of this?

cgzones:
No, that would probably require granting the LGTM.com GitHub-plugin some access on the repository.
But LGTM.com can analyse the project without any GitHub interaction just fine, only the analysis currently fails without a manual configuration, cause it is unable to detect the `autoconf-archive` dependency of the valgrind configure macros.

dburgener:
So in order to view the results of the analysis, we'd have to go to LGTM.com?

stevedlawrence:
If enabling a plugin in GitHub is needed to make this more useful, let me know and I can look into it. I'd prefer to to allow any kind of write access from LGTM though. Not sure what priviledges it requires.

cgzones:
Yes, I am unaware of any email notification system.

dburgener:
I'd say go for it @stevedlawrence.  Seems like it would be a useful addition to our CI pipeline.

dburgener:
I assume you meant *disallow* write access.  I agree with that.  On a quick scan of the docs, it looks like it only needs read though.

stevedlawrence:
Yep, i meant disallow. Agreed, permissions don't seem unreasonable. I've added the app to the selint repo. It think it should start showing up in new PR checks now.

PR #160, opened 2020-08-25
Leak test

cgzones:


dburgener:
@stevedlawrence ping

stevedlawrence:
Hmm, not sure why LGTM isn't showing up as a check, it's installed and has access as far as I can tell...Going to require a bit more digging I think.

dburgener:
I'm struggling to find it in any of repo settings.  Where should I be looking?

stevedlawrence:
You might not have permissions. When I go to this repos Settings > Integrations, it shows LGTM.com as an installed app. There's a button to "Configure", which says it has the approproriate priviledges for the ``TresysTechnology/selint`` repo. There are no other configuration options other that to Suspend or Uninstall the app.

dburgener:
Yeah, I don't have an "Integrations" menu under "Settings".  Is that something you can grant for me?

stevedlawrence:
I think it's only available for memebers of the TresysTechnology org, which I can't grant you access.

You could try adding LGTM.com to your fork, which you have full permissions for and see if you can get it working there. I must be missing something obvious.

dburgener:
Okay, thanks.  I'll look into adding it in my fork.

cgzones:
Maybe LGTM.com just needs some time; I'll retry in a couple days.
Unfortunately forks can not be analysed, only main repositories.
Thanks Steve for activating.

dburgener:
That makes sense.  I also don't see it showing up on LGTM.com yet.

stevedlawrence:
I've followed the directions on the [Automated code review FAQs](https://lgtm.com/help/lgtm/automated-code-review-faqs#how-can-i-tell-if-automated-code-review-is-enabled), and it says automated pull requests are enabled, so I think everything is configured correctly. I'd sugest creating a new PR and see if it's working now. It's also maybe possible that the "Draft" status of this PR is preventing triggering of LGTM, so maybe try not making it a draft if it still doesn't work.

dburgener:
I think I fixed it by modifying the config.  See https://github.com/TresysTechnology/selint/pull/163 where it appears to be working.

This one will probably work either as a new PR or if the checks are triggered again by activity on the branch.

dburgener:
If people are interested in seeing what you get from going to LGTM.com directly, you can see selint here: https://lgtm.com/projects/g/TresysTechnology/selint/alerts/?mode=list

Unfortunately the historical data, which I think would be one of the primary value adds outside of the PR checking, seems to all be failing because it's grabbing the lgtm config at the time of the commit, which is wrong.  So we only have usable data from the commit fixing the config forward.

dburgener:
Good to see it working.

PR #165, opened 2020-08-27
Allow xperm lists as macro argument

cgzones:
```
xperm_macro(foo_t, bar_t, 0x1234)
xperm_macro(foo_t, bar_t, { 0x5000 - 0x50ff 1234 })
```

### DIFF
--- a/src/string_list.h
+++ b/src/string_list.h
@@ -34,6 +34,10 @@ struct string_list *copy_string_list(const struct string_list *sl);
 struct string_list *sl_from_str(const char *string);
 struct string_list *sl_from_strn(const char *string, size_t len);
 
+// Return a string list with the given string as single item
+// Takes ownership of the given string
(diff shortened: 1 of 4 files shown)


src/string_list.h
```diff
@@ -34,6 +34,9 @@ struct string_list *copy_string_list(const struct string_list *sl);
 struct string_list *sl_from_str(const char *string);
 struct string_list *sl_from_strn(const char *string, size_t len);
 
+// Return a string list with the given string consumed
```
dburgener, line 37: Can the comment say something like "taking ownership of the string"?  I think that would help add clarity for someone who's confused by the term "consume" in the function name.

cgzones: Updated the function comment to:
```
// Return a string list with the given string as single item
// Takes ownership of the given string
```

PR #166, opened 2020-08-27
Add Codecov coverage check

cgzones:


dburgener:
Seems nice.  Once it has base data to compare to does it ever fail PRs for reducing coverage too much, or is it all just informational?

cgzones:
I think it only matters if Codecov has repository access (then a new check will appear), and it by default it'll fail if the covareg drops, but that can be configured, see https://docs.codecov.io/docs/commit-status

dburgener:
I think I like having this data available.  I don't want to fail PRs for reducing coverage though.  I'd of course prefer that contributors contribute tests, and I think that's a healthy open source model, but I don't want to discourage potential contributors, particularly new ones.  I'd rather look at the report and make a human decision on whether the PR is worth approving if it reduces coverage or if it needs more tests.

I've been running code coverage locally with gcov fairly often, but having it automated in PRs seems helpful.  If you want this and want to mark it as ready, I'm happy to approve it.

dburgener:
Does this need to be a separate check?  Any particular reason not to just run the gcc check with the --coverage flag on?

PR #167, opened 2020-08-31
Clone repositories for CI with depth 1

cgzones:
We only need a shallow checkout.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,28 +35,37 @@ jobs:
         sudo apt-get update -q
     - name: install dependencies
       run: sudo apt-get install -y uthash-dev libconfuse-dev autoconf-archive bison byacc flex valgrind help2man clang-10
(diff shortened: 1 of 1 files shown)


dburgener:
This doesn't look like much of a performance win to me.  On #166 check took 37s to run and bats took 0s.  On this, check took 36s to run and bats took 0s.  In the check case, since we're checking out a tag, I don't know that I see a downside, but in the bats case, I think this might run the risk of breaking things if they were to change their repo (which we should probably fix by going to a tag).  Am I missing some benefit here?

stevedlawrence:
You might consider using the github checkout action for check and bats repos. Presumably, github put some amount of thought into making that action as efficient/safe as possible. You can also specify tags/branches/etc. For example, I think the following is equivalent to your git checkout line:

```yaml
- uses: actions/checkout@v2
  with:
    repository: libcheck/check
    ref: 0.13.0
    path: check
```

It's a bit more verbose, but it lets github actions worry about the details.

cgzones:
Thanks Steve, that seems to work!

PR #169, opened 2020-08-31
Move coverage check after functional tests

dburgener:


dburgener:
@cgzones as soon as I hit merge on the coverage PR I noticed that you had it running before the functional tests, which are an important part of our coverage story.  This should help our coverage numbers more realistically reflect our actually coverage.  Look good?

cgzones:
Oops.
Thanks for noticing.
One question: do we want to include the coverage of the refpolicy parse test?
Strictly speaking we have no control over what policy constructs are getting scanned.

dburgener:
That's fair.  I was thinking that it's part of our CI so it's part of our overall coverage, but you're right that we get better consistency in the data if we don't have an external dependency like that.  I'll update to put it after "Test distribution" and before "checkout refpolicy".  Obviously "Test distribution" does nothing for coverage, but if we put it after then we can skip uploading coverage in the case where the distribution checks fail, since we don't really care about it at that point.

cgzones:
:+1:

PR #170, opened 2020-09-03
Allow the specification of multiple --context flags on the command line

dburgener:


src/main.c
```diff
@@ -465,8 +469,10 @@ int main(int argc, char **argv)
 
 	fts_close(ftsp);
 
-	if (context_path) {
-		paths[0] = context_path;
+	struct string_list *context_path_node = context_paths;
+
+	while (context_path_node) {
```
cgzones, line 474: I would use a for-loop, buts that just personal preference

dburgener: Yeah, I've noticed that.  :)  My preference is the while syntax, but I personally have no problems allowing both versions to coexist in the codebase.

src/string_list.c
```diff
@@ -121,6 +121,21 @@ struct string_list *concat_string_lists(struct string_list *head, struct string_
 	return head;
 }
 
+enum selint_error append_to_sl(struct string_list *sl, const char *string)
```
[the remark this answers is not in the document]
dburgener: Thanks.  Updated.

PR #175, opened 2020-10-08
Add check C-007 for unused self keyword

dburgener:


dburgener:
Looks like codecov decided to scan lex.c and parse.c this time which it hadn't before, which caused the coverage to appear to drop by 3%.  Not sure what the reason for that change is.

src/te_checks.c
```diff
@@ -140,6 +140,27 @@ struct check_result *check_unordered_perms(__attribute__((unused)) const struct
...
+	if (node->flavor != NODE_AV_RULE && node->flavor != NODE_XAV_RULE) {
+		return alloc_internal_error("Bad node type given to check C-007");
+	}
+	struct av_rule_data *av_data = node->data.av_data;
+
+	if (av_data->sources->next ||
+	    av_data->targets->next ||
+	    0 == strcmp(av_data->targets->string, "self") ||
+	    0 != strcmp(av_data->sources->string, av_data->targets->string) ||
+	    !look_up_in_decl_map(av_data->sources->string, DECL_TYPE) ) {
```
cgzones, line 155: Maybe also issue on attributes? `allow domain self:dir search_dir_perms;`

```
!look_up_in_decl_map(av_data->sources->string, DECL_TYPE)
```

->

```
(!look_up_in_decl_map(av_data->sources->string, DECL_TYPE) && !look_up_in_decl_map(av_data->sources->string, DECL_ATTRIBUTE))
```

cgzones: More specifically: What is the reason to exclude all non known identifiers, e.g. `$1` or `$1_t` ?

dburgener: The problem with attributes is that "allow attr attr" and "allow attr self" are not equivalent. If foo_t and bar_t both associate with attr, then "allow attr self" expands to "allow foo_t foo_t" and "allow bar_t bar_t", but it does not include "allow foo_t bar_t" or "allow bar_t foo_t". "allow attr attr" on the other hand does include the latter two.

The refpolicy variable case I hadn't thought of, but it would have the problem that we don't know if they are attributes or not.

Since the attribute mistake has actual consequences, it would be nice to detect it, but I'm not sure of a way to infer probable develop intent off the top of my head.

dburgener: $1_t would be reasonable. I can add that case, although I wonder how common such a case would be.

dburgener: I've added the $*_t.  Can you check and make sure I got the boolean logic right?

src/te_checks.c
```diff
@@ -140,6 +140,32 @@ struct check_result *check_unordered_perms(__attribute__((unused)) const struct
...
+	    0 == strcmp(av_data->targets->string, "self") ||
+	    0 != strcmp(av_data->sources->string, av_data->targets->string)) {
+		return NULL;
+	}
+
+	if (av_data->sources->string[0] != '$' &&
+	    !ends_with(av_data->sources->string, strlen(av_data->sources->string), "_t", 2) &&
+	     !look_up_in_decl_map(av_data->sources->string, DECL_TYPE)) {
+		return NULL;
+	}
```
cgzones, line 162: This will not skip `allow $1 $1:fd use;`

Maybe use
```c
if ((av_data->sources->string[0] != '$' ||
    !ends_with(av_data->sources->string, strlen(av_data->sources->string), "_t", 2)) &&
     !look_up_in_decl_map(av_data->sources->string, DECL_TYPE)) {
	return NULL;
}
```
?

dburgener: I'm getting overly confused with the complicated boolean expressions, so I've just gone with nested if/else for clarity.  Hopefully it's not just right now, it's also clearly right.

cgzones (approved):
LGTM :+1:

PR #176, opened 2020-12-02
Avoid recursive loop make_check_result() <-> alloc_internal_error()

cgzones:
Do not hang in case vasprintf() always returns < 0


dburgener:
Looks good.  Out of curiosity, did you hit this in the real world?

cgzones:
No; clang-tidy complained about it and I though this patch can't hurt.

dburgener:
Yeah totally reasonable.  I was just wondering if there was a real world use case where vasprintf was failing repeatedly and there weren't vastly bigger problems than reasonable shut down behavior on an SELinux policy static analysis tool.. :)

Thanks for the patch!

Issue #178, opened 2020-12-07
Comments followed by bare words in ifdefs no longer parse


dburgener:
Prior to a7f658f6c95e55a598bf8c0750aa16522d681b9e the following policy parsed successfully, now it does not:

```
ifdef(`foo',`
        # comment
        bare_word
')

```

The reason this could be valid policy is if bare_word is an m4_macro being expanded.  As a workaround, bare_word could be treated like an interface call as `bare_word()`.  I'm open to saying this is a wontfix if the consensus is that this isn't syntax we want/need to support, but the fact that we previously parsed it makes me lean towards parsing it.  I haven't had time to look into what the issue is or how easy it would be to fix yet.

PR #179, opened 2020-12-07
Accept bare word m4 macros

cgzones:
Accept m4 macros without any arguments.

Closes: #178

Downside:
Now we will accept all kind of artefacts.


src/parse.y
```diff
@@ -702,8 +708,6 @@ m4_argument:
 	|
 	BACKTICK lines SINGLE_QUOTE
 	|
-	BACKTICK strings SINGLE_QUOTE { free_string_list($2); }
```
dburgener, line 705: This seems concerning.  The commit message where this line was added mentions `define(`foo', `bar_t baz_t')` as something desirable to support.  My guess is that we'd now break on that?  It seems more desirable to me to support that then bare word macros.

dburgener: Or would those parse under the lines, since they'd be interpreted as bare_word macros?

dburgener:
Any idea why this worked before the multi-line change?  I don't remember the details of when I looked at it last, but I remember being somewhat confused as to what had changed.

dburgener:
> Downside:
> Now we will accept all kind of artefacts.

I'm super torn on this.  Part of me says that throwing a parse error on a random stray character is highly desirable and part of me says that we can't force policy developers to avoid bare word macros or else not use selint.  The general goal should be to minimize breakage when the policy compiles.  @cgzones what are your thoughts on this tradeoff?  @stevedlawrence, do you have any opinions here?

stevedlawrence:
Are bare word macros that common? I assume not since we would have since this before? If that's the case, what if we parse them as strings like this change does, but then add a "stray character check".  This way we error in the cases where a user really does have stray characters (which I agree seems pretty useful), but the check could still be disabled globally or with an "selint-disable" comment on lines where a user really does have a bare macro.

dburgener:
That sounds like a good idea to me @stevedlawrence.  Yes, I think they're very uncommon.  I noticed this because of one in a policy I'm working on, but it's the only time I've seen it.  In that case, it was a pretty trivial fix to change it from "bare_word" to "bare_word()" and get it parsing, so it's not a huge deal if we don't support it.  What bothers me more is that the policy was parsing correctly and then I updated selint and it broke.

I like the general idea of parsing it and checking for stray words.  I'll see if I can get such a check written in time for 1.2.0.

src/parse.y
cgzones: Yes, they would parse as `lines`, which is `lines line` and `line` can be a `STRING`.

cgzones:
I personally like SELint to be rather strict about unknown/uncommon constructs; to be warned early instead of finding out later a tool just hid suspicious code.

But I am fine moving this *strictness* from a hard parsing failure to a normal (ignorable) (error-)check.

dburgener:
This all looks good to me.  I'm ready to merge it if you want to mark it as ready.

dburgener:
@cgzones It looks like this broke this situation:

```
define(`not_foo',`-foo_t')
```

That's something we have in a policy I'm working on that I definitely need to support in my use case (moreso than the problem this PR addresses).

I tried adding "DASH STRING" into m4_simple_macro, but that shift/reduce conflicts with mls_range.  Any thoughts on a way to fix this problem?

My inclination is to say the supporting stray bare words may continue to cause problems in parsing and maybe the best call would be to revert this PR.  Thoughts?

cgzones:
m4 arguments are quite hard to identify and parse.

Is #187 a sound solution?

dburgener:
> m4 arguments are quite hard to identify and parse.

Very much so.  #187 seems reasonable at first blush, I think...  Let me take a little while to look/think more closely about it.

PR #182, opened 2020-12-10
Fix overflow on unexpected EOF

cgzones:


dburgener:
Any idea why LGTM is failing?  I don't see an issue in the checkout log it's pointing me to

dburgener:
Looks like it's failing on all PRs.  I'll merge this and see if that makes a difference for the others.

PR #187, opened 2020-12-15
Parse m4 define arguments as lax arbitrary string

cgzones:
Allows to parse
```
define(`not_foo',`-foo_t')
```

Note: with this patch the content of `defines` are not inserted into the syntax tree, e.g. with
```
define(`foo`, `type bar_t`)
```
`bar_t` is not added into the types-map.

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -677,6 +677,8 @@ m4_string_elem:
 	COLON
 	|
 	SEMICOLON
+	|
+	DASH
 	;
 
 condition:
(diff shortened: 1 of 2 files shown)


dburgener:
My central concern here is that we won't have information available if we decide to parse things like:

```
define(`getattr_dirs_pattern',`
        allow $1 $2:dir search_dir_perms;
        allow $1 $3:dir getattr_dir_perms;
')
```

I do think at the end of the day, maybe that's okay.  My inclination is to take this patch so that we can parse arbitrary defines, which seems valuable.

dburgener:
I'm ready to take this assuming you're okay with it.  Do you want to mark it as ready for review?

IMO we should take this, and then rebase #189 on top of it.

PR #188, opened 2020-12-15
 Adding NULL check around res before attempting to free res->message

gdesrosi:
Adding NULL check around res before attempting to free res->message

### DIFF
--- a/src/check_hooks.c
+++ b/src/check_hooks.c
@@ -294,7 +294,9 @@ void display_check_issue_counts(const struct checks *ck)
 
 void free_check_result(struct check_result *res)
 {
-	free(res->message);
+	if (res) {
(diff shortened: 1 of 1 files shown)


dburgener:
Github doesn't seem to recognize that you're the author of the commit (it's not connecting your e-mail from the commit with your github account).  That's not actually a problem from an SELint perspective, but it won't show up as a contribution on your github profile.  I suspect the right answer is to add the e-mail associated with this commit to your github account, although you could also change this commit author to match your github account.

gdesrosi:
@dburgener I just updated my account to use the new owl email address hopefully that fixed it.

dburgener:
Yep, looks good.  Thanks for the contribution!

PR #189, opened 2020-12-16
Fix E-010 false positives in define() macro calls 

WOnder93:
Bare words in the first argument to define() are perfectly OK, as they name a newly defined macro. Therefore ignore them in the bare word check.


dburgener:
Thanks!  Seems reasonable to me.  @cgzones does this seem like it will conflict with #187 in any way?  Obviously whichever one gets taken second will need a rebase, but my reading is that they should be able to coexist just fine.  Do you agree?

cgzones:
If #187 gets accepted, the second commit gets obsolete:

Currently the first *name* argument of a `define` statement gets parsed as `m4_args`
(which can be parsed as `m4_args` -> `m4_argument` -> `lines` -> `line` -> `bare_line` -> `m4_simple_macro` and thus lead to an E-010 tag)
https://github.com/TresysTechnology/selint/blob/8723a68ff0492f62454afe5e1f6215ab871314e7/src/parse.y#L841-L843

After #187 the *name* argument of a `define` statement gets parsed as `define_name`, which can now only be fulfilled by a (optional m4-backtick-singlequote-quoted) string.
So the *name* can never get parsed as a `m4_simple_macro` and so not anymore lead to an E-010 tag.

```flex
define:
	DEFINE OPEN_PAREN define_name COMMA define_expansion CLOSE_PAREN
	;

define_name:
	BACKTICK STRING SINGLE_QUOTE { free($2); }
	|
	STRING { free($1); }
	;
```

p.s.: this is mostly a side-effect of #187, which purpose is mainly to ignore the second *expansion* argument of `define` statements.

dburgener:
@WOnder93 @cgzones I've rebased this now that #187 merged.  Per Christian's suggestion I dropped the functional portion of the second commit as now obsolete, but kept the additional test code (although because of #187 it now hits one fewer time in e10.warn.te, which I've changed the test driver to reflect in an additional commit with an explanation).

Does what I've done seem reasonable?  Any objections to merging this as it stands now?

cgzones:
LGTM :+1:

WOnder93:
I don't know... now the commits look kind of awkward. The last commit should probably be squashed into the previous one (it just undoes a part of it) and the commit message of the second one doesn't really describe what it does any more (now it just adds the test). And the first commit is probably useless now, since there is no user of the DEFINE block being added any more.

dburgener:
That's fair.  I suspect there's some sort of long term value in inserting the defines into the syntax tree in case people want to check on them in the future.  Would you feel comfortable with just squashing your two commits together with the message from your first commit?

WOnder93:
@dburgener I squashed the commits and updated the commit message to better reflect the changes. I kept the first commit as you requested. Is it OK like this?

dburgener:
Oops, I got myself confused, and meant to take this prior to 1.2.0.  Sorry!  Since the actual underlying bug was fixed, I guess it's not urgent enough to warrant an immediate 1.2.1.  I'll plan on merging this shortly though, as I think it looks good now.

PR #190, opened 2020-12-20
add specfile for building selint rpm

dschadlich:
Found that there was no specfile for building an rpm. I have tested this specfile targeting el7, el8 and f33.

### DIFF
--- a/selint.spec
+++ b/selint.spec
@@ -1,0 +1,47 @@
+Summary: SELinux policy source file checker
+Name: selint
+Version: 1.1.0
+Release: 1%{?dist}
+URL: https://github.com/TresysTechnology/selint
+License: ASL 2.0
+%global forgeurl https://github.com/TresysTechnology/selint
+%global branch master
+# handle not having forge macros on el7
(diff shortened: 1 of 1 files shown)


dburgener:
Thanks for your contribution!  Sorry about the delay in responding over the holidays.

This all looks pretty reasonable to me, although my main question is that I'm not aware of many (any?) projects that carry their own spec files as part of their source repos.  As I understand it, that is typically done by the distros.  Can you help me understand why this should be included in SELint, rather than handled by an rpm based distro that wishes to include it?

dschadlich:
No problem, hopefully you had an enjoyable holidays.

I have seen a few projects that have their own spec files, I could looks for some examples, direwolf is one that comes to mind right away. As for if the specfile should even be in the repo, I have mixed feelings. Ideally it would be handled by the distro (fedora), and they would maintain it. That however does not help people who are on a distros that do regularly get new packages (like EL7 or EL8). Having the specfile in the repo would enable people on these distros to build an rpm and install it. In my particular use case is trying to install and use SELint on EL7. I am going to be making efforts to try to get SELint included within Fedora, but that is a separate effort.

dburgener:
Merged.  Thanks for the PR!

PR #195, opened 2021-01-19
Fix 32-bit sign comparisons

cgzones:
```
parse.y: In function 'yyerror':
parse.y:1111:26: error: comparison of integer expressions of different signedness: 'int' and 'unsigned int' [-Werror=sign-compare]
 1111 |     if (c - current_line < current_first_column) {
      |                          ^
parse.y:1113:33: error: comparison of integer expressions of different signedness: 'int' and 'unsigned int' [-Werror=sign-compare]
 1113 |     } else if (c - current_line < current_last_column) {
      |                                 ^
```


dburgener:
Merged, thanks

PR #197, opened 2021-01-19
Always include man/selint.h2m in dist-archive

cgzones:


dburgener:
This seems good to me.

dburgener:
@cgzones Do you want to mark this as ready for review?

PR #198, opened 2021-01-25
Allow define() without expansion

WOnder93:
This allows parsing statements such as "define(`something')", which are
perfectly valid.


dburgener:
Looks great, thanks!

PR #201, opened 2021-04-27
Clang 12 and CI update

cgzones:
* Modernize CI to focal (20.04)
  - use clang-12
  - use packaged bats, it is recent enough
  - update check version to 0.15.2
  - drop no more needed compiler warning overrides for code generated by
    old flex/bison versions

* Ignore Clang 12 UBSAN reports in hash functions
  Clang 12 complains about unsigned-shift-base in `HASH_FIND()`.


dburgener:
Wonderful!  Thanks.

PR #202, opened 2021-04-27
RFC:  Add extra check for unused interfaces

cgzones:
Disabled by default, needs explicit activation.

Add a new severity level 'X' for this introspection check.


dburgener:
Looks good to me.  Go ahead and mark as ready if you're ready for it to get merged.

PR #208, opened 2021-05-31
Allow { in defines in .te files

blenk92:
Currently
```
define(`something', `{ thing1 thing2 }')
```
leads to an parser error because the curly braces are not allowed in the
second argument of the definition. As this seems to work in practise,
this seems like a parser problem to me. This commit allow the curly braces
to be used in these definitions.


dburgener:
Looks great, thanks!

PR #209, opened 2021-07-02
Check attribute names in declarations

cgzones:
For declarations in interfaces like

    type foo_t, bar_attr;

also include in checks W-001 and W-002 the attribute bar_attr.

See https://github.com/SELinuxProject/refpolicy/pull/384

ssh.if:             172: (W): Attribute ssh_server is used in interface but not required (W-002)
W-002: 1


dburgener:
Merged.  Thanks!

PR #211, opened 2021-07-02
Add checks regarding conditional expression identifiers

cgzones:
Check each identifier is known.
Check each identifier is declared in own module or globally.

Refpolicy findings:

    mplayer.te:         122: (C): Identifier xserver_allow_dri in expression for conditional block not found in own module, but in module xserver (candidate for global declaration or interface) (C-008)
    xguest.te:           44: (C): Identifier user_exec_noexattrfile in expression for conditional block not found in own module, but in module userdomain (candidate for global declaration or interface) (C-008)
    xguest.te:           48: (C): Identifier user_rw_noexattrfile in expression for conditional block not found in own module, but in module userdomain (candidate for global declaration or interface) (C-008)
    nscd.te:            125: (C): Identifier samba_domain_controller in expression for conditional block not found in own module, but in module samba (candidate for global declaration or interface) (C-008)
    usermanage.te:      578: (C): Identifier samba_domain_controller in expression for conditional block not found in own module, but in module samba (candidate for global declaration or interface) (C-008)
    userdomain.if:     1236: (C): Identifier usbguard_user_modify_rule_files in expression for conditional block not found in own module, but in module usbguard (candidate for global declaration or interface) (C-008)
    C-008: 6

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -210,6 +210,10 @@ selinux_file:
 	|
 	comments av_file
 	|
+	global_conditions_file
+	|
+	comments global_conditions_file
+	|
 	comments
 	;
 
@@ -677,7 +681,7 @@ m4_string_elem:
 	;
 
 condition:
-	STRING { free($1); }
+	STRING { save_identifier(cur->parent, $1); }
 	|
 	NOT condition
 	|
@@ -1058,6 +1062,34 @@ av_permission:
 	STRING { insert_into_decl_map($1, "__av_file__", DECL_PERM); free($1); }
 	|
 	COMMENT
+	;
+
+// policy/global_booleans and policy/global_tunables files
+
+global_conditions_file: // must not allow a comment to be first -> parser conflict
+	gcf_definition gcf_contents
+	|
+	gcf_definition
+	;
+
+gcf_contents:
+	gcf_contents gcf_content
+	|
+	gcf_content
+	;
+
+gcf_content:
+	gcf_definition
+	|
+	COMMENT
+	;
+
+gcf_definition:
+	bool_declaration {
+		if (expected_node_flavor != NODE_COND_FILE) {
+				const struct location loc = { @1.first_line, @1.first_column, @1.last_line, @1.last_column };
+				yyerror(&loc, NULL, "Error: Unexpected global conditionals file parsed"); YYERROR;
+		} }
 	;
 
 %%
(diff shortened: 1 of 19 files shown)


src/parse.y
```diff
@@ -1071,6 +1075,34 @@ av_permission:
 	COMMENT
 	;
 
+// policy/global_booleans and policy/global_tunables files
+
+global_conditions_file: // must not allow a comment to be first -> parser conflict
```
dburgener, line 1080: In default refpolicy both of these files do start with comments.  What is the impact of that on this?  Would we be adding an error message out of the box for anyone trying to run selint on vanilla refpolicy?

src/te_checks.c
```diff
@@ -663,6 +701,77 @@ struct check_result *check_unknown_interface_call(__attribute__((unused)) const
 				 if_name);
 }
 
+static bool is_bare_parameter(const char *name)
```
dburgener, line 704: Is "bare parameter" a common term that I'm just not aware of?  I had to read the code pretty closely to figure out what this was actually checking.  So unless we expect policy developers to know/use that term and I'm just the one out of the loop here, can we add a comment defining "bare parameter" above this function?

src/parse.y
cgzones: It's similar to the comment at https://github.com/TresysTechnology/selint/blob/c4cc6e09b8c66fa920c8ebc1e942eac3f554121b/src/parse.y#L993

Leading comment in the file are allowed and accepted, the comment is just about not allowing leading comments in the 
 `global_conditions_file` symbol, cause they are handled in the symbol `selinux_file` via the rule
```
        [...]
        global_conditions_file
	|
	comments global_conditions_file
        [...]
```

src/te_checks.c
cgzones: Added a comment to the function, but I am also happy to rename it; suggestions?

dburgener: If there's a commonly accepted term for the distinction you're making here, I'm not sure what it is.  I do like "simple" as used in your comment better than "bare" I think.  I don't know that I have a better suggestion.  I think the example in the comment clarifies the intent quite nicely.

src/parse.y
dburgener: Thanks, I understand now.

PR #215, opened 2021-11-05
Miscellaneous parser simplifications

cgzones:


dburgener:
I'm a little skeptical of 9e60f532edb46a48ff1680f1b3c2ec392e0508b4.  I get that this is unused now.  But I feel like if we want to capture an optional string in a macro at some point in the future it will come in handy as written.  I'm not sure the case for the simplication/optimization now is really worth it in terms of future potential use as the code base evolves?  I'm very open to push-back if you disagree though.

dburgener:
Really the same comment for 1a9d02ba87b6ef2d7e57baabb5a81cea25050371.  Both of these situations seem to me like a philosophical question between optimizing the code base for its current state where these things are unused vs premptively preparing for hypothetical future uses.  I'm not sure I see a particularly strong benefit to removing the unused code in these places.

dburgener:
The rest of the commits look great.  Thank you very much for looking into this!

cgzones:
Dropped the two commits in question.

dburgener:
Merged.  Thanks!

PR #216, opened 2021-11-30
Support IPv4/IPv6 address embedding

cgzones:
Accept IPv4 addresses embedded in IPv6, like `::ffff:127.0.0.1`.
This allows using those in nodecon statements.


dburgener:
I had no idea this was a thing you could do with ipv6.  It's kind of horrific in every way, but since it's a thing, I guess we need to support it.  Thanks!

cgzones:
It's not supported by libsepol yet, but I hope to introduce it: https://patchwork.kernel.org/project/selinux/patch/20211130110034.12920-1-cgzones@googlemail.com/

PR #217, opened 2022-01-06
Support policy_module macro without version

cgzones:
Refpolicy dropped the version information from the policy_module macro
and just uses the macro now with only the module name.

https://github.com/SELinuxProject/refpolicy/commit/78276fc43b76277399f38028b4a2904b3b541e6c
https://github.com/SELinuxProject/refpolicy/pull/456

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -237,10 +237,10 @@ maybe_selint_disable:
 
 
 header:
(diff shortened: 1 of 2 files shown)


dburgener:
Merged. Thanks!

PR #219, opened 2022-01-07
Update Changelog and version info for 1.2.1

dburgener:


dburgener:
I need to keep notes on the correct way to make a release...  Do I have to push a tag directly to master?  @stevedlawrence I think you set up the release CI.  Can you remind me the right way to do it?

stevedlawrence:
Yep, just push a tag that starts with `v` to master and everything else is automated.

stevedlawrence:
Your changelog changes mention 1.3.0, should be that be 1.2.1?

dburgener:
Do I need to tag a new commit to trigger the CI, or can I just push a tag to the existing HEAD?

CHANGELOG
```diff
@@ -1,5 +1,15 @@
 # SELint Changelog
 
+## [1.3.0] 2022-01-10
```
stevedlawrence, line 3: Should be 1.2.1?

dburgener: fixed, thanks

stevedlawrence:
You just have to push the tag, it's fine if the commit already exists (and probably preferred so you get one last run of the GtiHub actions before officially tagging this). A push of any kind, whether it be a tag or a commit triggers the workflow.

dburgener:
Awesome, thanks.

PR #220, opened 2022-01-26
Check for argument counter of zero

cgzones:
Reject invocations with an argument counter of zero, in which case the
subsequent allocation

    char **paths = malloc(sizeof(char *) * (unsigned)argc - (unsigned)optind + 2);

will cause a crash in fts_read(3).


dburgener:
Merged, thanks!

PR #223, opened 2022-02-09
CI: bump to LLVM version 13 and really use the desired Clang version

cgzones:
Currently the Clang build uses default installed Clang version (11),
since the additional installed package does not override the default
command without a version suffix.


dburgener:
Merged.  Thanks!

PR #226, opened 2022-03-30
Add check for allow rules with excluded source or target

cgzones:
Excluding a type/attribute from an attribute as the source or target in
an allow statement generates av rules for each non excluded type.  If
the attribute has quite an amount of types assigned those statements can
result in many rules and a binary policy size increase.


dburgener:
I love it.  I've encountered these same concerns around binary size recently as well.

I wonder if there's a way to document the binary size motivation better?  I imagine users would be confused as to why they would want this, and the git commit message is the only way the motive is explained now.

An extra sentence in the comment above the check declaration would help.  I wonder if a brief note in the README like "(can affect policy binary size)" would be appropriate as well?  What are your thoughts?

cgzones:
> An extra sentence in the comment above the check declaration would help. I wonder if a brief note in the README like "(can affect policy binary size)" would be appropriate as well? What are your thoughts?

Agree.

It seems to also need some improvements as I like to catch statements like:

```
manage_files_pattern(foo_t, file_type, { file_type -secret_t })
```

dburgener:
Merged, thanks!

PR #228, opened 2022-04-14
Update naming related to X-002

cgzones:
The check X-002 also scans dontaudit and auditallow rules.

Follow up of #226


dburgener:
Good catch.  Merged, thanks.

PR #234, opened 2022-06-28
Avoid exceptions override

cgzones:
In a require block of

    gen_require(`
        type a; #selint-disable: W-010
        type b; #selint-disable: W-011
    ')

the SELint exceptions "W-011" will override "W-010" of the declaration
for type a and leak the previous one set.

Fixes: da6d80e7 ("Apply command to all declarations in a single require statement")

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -578,40 +578,57 @@ require_lines:
 	;
 
 require_line:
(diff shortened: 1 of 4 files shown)


dburgener:
Merged, thanks!

Issue #232, opened 2022-06-26
selint -r should show full path to files in error messages


joenall:
If I invoke `selint -r .` at the top of a deep tree, it would be very helpful if the error message contained the relative or full path to the file in error. I have 329 .te files in 6,004 directories.
I tried to get around this with a:
```
find . -name \*.te | xargs selint
```
but the path is still trimmed.

PR #235, opened 2022-06-28
Print full paths

cgzones:
* Display full path on parser errors
  Parser error are severe and also quite rare.  Print the full path by default.

* Add option to print full paths
  Add command line option '--full-path' to print the full path for issues.

Closes: #232


dburgener:
Can we add a test for the full-path flag?  That seems like something that would be easy to break and not notice.

dburgener:
Wonderful, thanks!

PR #237, opened 2022-06-28
Check memory allocations

cgzones:
Check all allocations for failures (OOM), not only to avoid mostly
harmless NULL-dereferences, but also UB and memory corruption caused by
optimizations.

See also: https://pvs-studio.com/en/blog/posts/cpp/0938/

Occurrences:
    grep -REw "(malloc|calloc|strdup|strndup|realloc|reallocarray)" src/

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 bin_PROGRAMS = selint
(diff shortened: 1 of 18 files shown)


dburgener:
I found the article mostly unconvincing, but I guess "We should allow undefined behavior" isn't a hill I want to die on, and the provided approach is pretty clean, so I'll take this.  Thanks!

Issue #230, opened 2022-06-25
Failure to parse gen_require all_userspace_class_perms


joenall:
This:
```
gen_require(`
        all_userspace_class_perms
'
```
causes selint to emit:
```
$ selint -r .
Note: Check E-007 is not performed because no permission macro has been parsed.
sepgsql-regtest.te:   4: (F): syntax error, unexpected STRING (F-001)
    4 |     all_userspace_class_perms
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~
sepgsql-regtest.te:   3: (F): Error: Invalid statement (F-001)
    3 | gen_require(`
      | ^~~~~~~~~~~~~
    4 |     all_userspace_class_perms
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~
```
The code dates back to [pgsql](https://marc.info/?l=pgsql-hackers&m=130198501823082&w=2) and all_userspace_class_perms is in /usr/share/selinux/devel/include/support/all_perms.spt

If I change the gen_require to:
```
gen_require(`
        class db_database all_db_database_perms;
        class db_table all_db_table_perms;
        class db_procedure all_db_procedure_perms;
        class db_column all_db_column_perms;
        class db_tuple all_db_tuple_perms;
        class db_blob all_db_blob_perms;
        class db_schema all_db_schema_perms;
        class db_view all_db_view_perms;
        class db_sequence all_db_sequence_perms;
        class db_language all_db_language_perms;
')
```
the fatal error goes away.

dburgener:
This should be fixed by #238.

PR #238, opened 2022-07-27
Parse simple m4 macros in gen_require statements

dburgener:


dburgener:
Closes #230.  This only addresses parsing.  SELint still wouldn't know that the userspace object classes and perms are actually included in this situation.  I think that's okay in the current state, because I don't think we actually do any linting that needs to know the specific classes and perms.  As far as I see, our lints on requires are limited to types, attributes and roles.

PR #239, opened 2022-08-15
Provide strict prototypes in function definitions

cgzones:
Clang 15 starts to complain about non-strict prototypes in function
definitions:

    parse_functions.c:67:30: warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]
    char *get_current_module_name()
                                 ^
                                  void


dburgener:
Looks good, thanks!  Will merge once checks complete assuming no issues.

PR #242, opened 2022-08-18
Don't include interface calls to templates in template declarations

### LINKED ISSUE #241: Erroneous W-002 in template
I am seeing W-002 findings inside templates that I believe are erroneous.  I am declaring a type $1_t and any time I use that type in the template it generates a W-002.  If I change the type to $1_dummy_t the W-002 findings are no longer generated, but $1_t declaration is still valid.

I created a small dummy.if file attached in a zip.
[dummy.zip](https://github.com/SELinuxProject/selint/files/9365447/dummy.zip)

selint command:
`selint -s --context=./policy ./policy/< path to dummy.if >`

dburgener:
When an interface calls a template, this is invalid interface usage, and
generates an S-004 result.  However, if that interface calls templates,
it will not substitute the interface arguments in the template call,
resulting in the template map thinking that types named $1_t and similar
are valid types.  This can then trigger W-002 in templates.

Resolving the S-004 result also removes the W-002 warnings, but the
false positives can be confusing for users.

By not inserting interfaces calls in the template declarations, we avoid
treating types declared in interfaces as valid, which silences these
W-002 results.  These declarations were weird anyways, since they were
associated with the module of the interface rather than the module of
the te file calling the interface.

closes #241


tests/functional/policies/check_triggers/w02.bad_if.if
```diff
@@ -0,0 +1,37 @@
...
+##      </summary>
+## </param>
+#
+template(`dummy_template',`
+        gen_require(`
+                type etc_t;
+        ')
+
+        type $1_t;
+        domain_type($1_t) # Generates a W-002
```
cgzones, line 20: This (and also the next) comment is a bit misleading, since the test runner `test_one_check_expect "W-002" "w02.bad_if.if" 0` expect zero warnings.

dburgener: Thanks!  Yeah, I copied the file the issue reported supplied for the test and it had his comments from when they were reporting those warnings.  Should be fixed now.

PR #247, opened 2022-09-19
Add new check to warn about incorrect usage of audit_access permission

cgzones:
The audit_access permissions for the common file class has only an
affect in dontaudit rules (and to an extend in neverallow rules).
Warn about occurrences in allow and auditallow rules.

Refpolicy findings (due to usage in unconfined access):

    kernel.te:          551: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          552: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          553: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          555: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          556: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          560: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          561: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          562: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          563: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          564: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          565: (W): Allow rule with audit_access permission (W-013)
    kernel.te:          566: (W): Allow rule with audit_access permission (W-013)
    filesystem.te:      341: (W): Allow rule with audit_access permission (W-013)
    filesystem.te:      342: (W): Allow rule with audit_access permission (W-013)
    filesystem.te:      343: (W): Allow rule with audit_access permission (W-013)
    filesystem.te:      344: (W): Allow rule with audit_access permission (W-013)
    filesystem.te:      345: (W): Allow rule with audit_access permission (W-013)
    filesystem.te:      346: (W): Allow rule with audit_access permission (W-013)
    filesystem.te:      347: (W): Allow rule with audit_access permission (W-013)
    storage.te:          62: (W): Allow rule with audit_access permission (W-013)
    storage.te:          63: (W): Allow rule with audit_access permission (W-013)
    devices.te:         437: (W): Allow rule with audit_access permission (W-013)
    devices.te:         438: (W): Allow rule with audit_access permission (W-013)
    devices.te:         439: (W): Allow rule with audit_access permission (W-013)
    files.te:           230: (W): Allow rule with audit_access permission (W-013)
    files.te:           231: (W): Allow rule with audit_access permission (W-013)
    files.te:           232: (W): Allow rule with audit_access permission (W-013)
    files.te:           233: (W): Allow rule with audit_access permission (W-013)
    files.te:           234: (W): Allow rule with audit_access permission (W-013)
    files.te:           235: (W): Allow rule with audit_access permission (W-013)
    files.te:           236: (W): Allow rule with audit_access permission (W-013)
    W-013: 31

### DIFF
--- a/tests/functional/policies/check_triggers/w13.te
+++ b/tests/functional/policies/check_triggers/w13.te
@@ -1,0 +1,13 @@
+policy_module(w13, 1.0)
+
+type foo_t;
+
+# do not warn about these
+allow foo_t foo_t:file *;
+allow foo_t foo_t:dir ~map;
+dontaudit foo_t foo_t:chr_file audit_access;
+neverallow foo_t foo_t:sock_file audit_access;
+
+# warn about these
+allow foo_t foo_t:file audit_access;
+auditallow foo_t foo_t:lnk_file audit_access;
--- a/README
+++ b/README
@@ -187,6 +187,7 @@ CHECK IDS
 	W-010: Call to unknown interface
 	W-011: Declaration in require block not defined in own module
 	W-012: Conditional expression contains unknown identifier
+	W-013: Incorrect usage of audit_access permission
 
 	E-002: Bad file context format
 	E-003: Nonexistent user listed in fc file
--- a/check_examples.txt
+++ b/check_examples.txt
@@ -148,6 +148,13 @@ W-011:
 		')
 	')
 
+W-012:
+
+	# in module foo
+	if (bar_cond) {
+		...
+	}
+
 Error:
 
 E-002:
--- a/check_examples.txt
+++ b/check_examples.txt
@@ -155,6 +155,10 @@ W-012:
 		...
 	}
 
+W-013:
+
+	allow foo_t bar_t:file audit_access;
+
 Error:
 
 E-002:
--- a/src/check_hooks.h
+++ b/src/check_hooks.h
@@ -64,6 +64,7 @@ enum warn_ids {
 	W_ID_UNKNOWN_CALL      = 10,
 	W_ID_IF_DECL_NOT_OWN   = 11,
 	W_ID_UNKNOWN_COND_ID   = 12,
+	W_ID_AUDIT_ACCESS      = 13,
 	W_END
 };
 
--- a/src/runner.c
+++ b/src/runner.c
@@ -281,6 +281,10 @@ struct checks *register_checks(char level,
 			add_check(NODE_TUNABLE_POLICY, ck, "W-012",
 				  check_unknown_cond_id);
 		}
+		if (CHECK_ENABLED("W-013")) {
+			add_check(NODE_AV_RULE, ck, "W-013",
+				  check_audit_access_perm);
+		}
 		// FALLTHRU
 	case 'E':
 		if (CHECK_ENABLED("E-002")) {
--- a/src/te_checks.c
+++ b/src/te_checks.c
@@ -830,6 +830,26 @@ struct check_result *check_unknown_cond_id(__attribute__((unused)) const struct 
 	return NULL;
 }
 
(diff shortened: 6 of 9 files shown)


tests/functional/policies/check_triggers/w13.te
```diff
@@ -0,0 +1,13 @@
...
+type foo_t;
+
+# do not warn about these
+allow foo_t foo_t:file *;
+allow foo_t foo_t:file ~map;
+dontaudit foo_t foo_t:file audit_access;
+neverallow foo_t foo_t:file audit_access;
+
+# warn about these
+allow foo_t foo_t:file audit_access;
```
dburgener, line 12: Just for the sake of thoroughness, maybe add a check for another object class, such as dir as well?

dburgener:
Calling this a "warning" seems questionable to me, since:

1. Every finding in current refpolicy seems incredibly innocuous.
2. Allowing access here is a noop.

The question is what to call it.  I do see the positive case for warning (broadly, "you presumably wrote this intending to do something, but it does nothing" sounds like a "warning".  It's just that the existing examples in refpolicy seem to me to make a positive case for inclusion of audit_access, and it's hard to call something that seems legitimate a "warning".

I think a question that could be helpful is: If we take this check, how would we modify refpolicy for passing?  These commands could be changed to use "*", we could add a disable comment, we could drop audit_access, or we could disable the check entirely.  I think dropping audit_access actually seems undesirable in this instance, since having an "unconfined" that's missing permissions is something that someone might look at and think is wrong.  Better to just avoid the question, since it's a noop.  

"*" seems preferable to me so that any future permissions are included, since these are unconfined.  Although the lack of * now may indicate a desire by Chris to keep these lists explicit, in case some problematic new permission he didn't want to allow was added to the kernel.

Both disables are of course workable solutions, although creating a check just to disable it in refpolicy seems like a case that the check shouldn't be a "warning".

I think I'm talking myself into wanting to call this "style" rather than warning.  I'm very open to other suggestions, pushback, and thoughts about the "how to address the check in refpolicy" angle.

cgzones:
Initially I choose severity warning cause it's not so much about a question of style (how do I handle access(2) requests (allow, dontaudit, dontaudit with audit_access)), but more about a statement, which is from the access decision side is a noop, but probably might not be indented that way and might do the wrong thing (maybe the author wanted to allow the access(2) request, maybe he wanted to dontaudit it).

My way of handling the findings in Refpolicy would be to drop the permissions from the rules, since that would not change the behavior of the policy. Using the wildcard `*` permissions would add some permissions (`watch_mount`, `watch_sb`, `watch_with_perm`, `watch_reads` and sometimes `map`) and clash with check W-008. An alternative might be to introduce `unconfined_*_perms`, so added permissions in the future by the kernel need to be added only in one place.

@pebenito what is your opinion on the current usage of `audit_access` in allow rules, e.g. https://github.com/SELinuxProject/refpolicy/blob/abb3474508a94b7b704ac8cfcd88d0e4971d135d/policy/modules/kernel/kernel.te#L551, and the "severity" of them being a noop?

pebenito:
Since the only misuse of the perm is in the unconfined rules, I'd prefer to drop the permission. It seems like the default for selint should notify the user but be nonfatal. For refpolicy I would want it to be fatal.

pebenito:
SELinuxProject/refpolicy#543

tests/functional/policies/check_triggers/w13.te
cgzones: done

dburgener:
Okay, I'm okay with calling this a warning.  Will merge.  Thanks!

PR #248, opened 2022-09-19
Update repository URLs and use https license links

cgzones:


LICENSE
```diff
@@ -1,7 +1,7 @@
 
                                  Apache License
                            Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                        https://www.apache.org/licenses/
```
dburgener, line 4: I think modifying the license text, however trivially, seems legally unwise.  See the Apache FAQ: https://www.apache.org/foundation/license-faq.html#mod-license

As far as I can tell, the headers with the license links are not legally binding in the same way, and modifications to them that don't change the core meaning seem less concerning.  Not to mention that Apache's website seems to have updated to https themselves there (https://www.apache.org/foundation/license-faq.html#Apply-My-Software).  Note that the license text at https://www.apache.org/licenses/LICENSE-2.0 still uses http, I assume for exactly the same reason that modifying the license text at all is legally risky.

So I'd be fine with the other http -> https changes, but I won't take any changes to the license text itself.

cgzones: True, I only verified the link works and saw the https link mentioned at that page in the `Text version` portion, but not the actual license.
Will drop that commit.

stevedlawrence: ASF has actually discussed this. I believe the decision they came up with is that it's not necessary to change to https, but it's legally fine to use https if you want, and they leave it up to individual projects to decide what they want to do.

I believe this is the issue where they had this discussion:

https://issues.apache.org/jira/browse/LEGAL-457

dburgener: Thanks for the link, Steve!  I continue to remain more comfortable maintaining the text exactly identical to 1. How we have always had it and 2. How Apache has it, and avoid any concerns about "relicensing", or whether a "modified" version of the license is still "identical" to the Apache license.

cgzones: Dropped the commit in question.

dburgener:
Merged, thanks!

PR #249, opened 2022-09-19
Misc Clang findings

cgzones:


dburgener (approved):

dburgener:
The changes look good, thanks.

I take it we haven't seen these before in the CI because we don't have -Werror enabled on clang builds?  Does this PR allow us to enable that on clang?  Is that something we want?

cgzones:
We are already using `-Werror` in the CI, just not Clang 15:

https://github.com/SELinuxProject/selint/blob/39aa318d77ba404266f008bfdd38c7698c81abb2/.github/workflows/main.yml#L64-L71

PR #250, opened 2022-09-20
RFC: Expand checks about requirements to userspace classes

cgzones:
Expand the check `W-001`, `W-002` and `W-003` to security classes.

Since the `system` class is used both by the kernel and in userspace by systemd some machinery needs to be added to gather the associated used permission.
This can also be reused to infer the type of identifiers, see #206.

### DIFF
--- a/src/if_checks.c
+++ b/src/if_checks.c
@@ -262,7 +262,7 @@ struct check_result *check_name_used_but_not_required_in_if(const struct
 
 	const struct policy_node *cur = node;
 
-	struct string_list *names_in_current_node = get_names_in_node(node);
+	struct name_list *names_in_current_node = get_names_in_node(node);
 
 	if (!names_in_current_node) {
 		return NULL;
@@ -276,15 +276,15 @@ struct check_result *check_name_used_but_not_required_in_if(const struct
 	}
 
 	if (!cur) {
-		free_string_list(names_in_current_node);
+		free_name_list(names_in_current_node);
 		return NULL;
 	}
 	// In a template or interface, and cur is a pointer to the definition node
 
 	cur = cur->first_child;
 
-	struct string_list *names_required = NULL;
-	struct string_list *names_required_tail = NULL;
+	struct name_list *names_required = NULL;
+	struct name_list *names_required_tail = NULL;
 
 	while (cur && cur != node) {
 	       if (cur->flavor == NODE_GEN_REQ
@@ -305,7 +305,7 @@ struct check_result *check_name_used_but_not_required_in_if(const struct
 		                     // for example in an ifdef
 	}
 
-	const struct string_list *name_node = names_in_current_node;
+	const struct name_list *name_node = names_in_current_node;
 	/* In declarations skip the first name, which is the new declared type */
 	if (node->flavor == NODE_DECL) {
 		name_node = name_node->next;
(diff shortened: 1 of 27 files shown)


dburgener:
Just a quick comment on the RFC nature of this: The high level idea seems really valuable, and like a good addition to SELint.  Thanks for putting in the work on this!

Looks like a pretty large chunk of code, so I don't think I'll have time to review it this week, but I'll aim to set aside some time next week.

src/if_checks.c
```diff
@@ -330,6 +330,13 @@ struct check_result *check_name_used_but_not_required_in_if(const struct
 				flavor = "Role Attribute";
 			} else if (name_is_role(ndata) && look_up_in_decl_map(ndata->name, DECL_ROLE)) {
 				flavor = "Role";
+			} else if (name_is_class(ndata) && look_up_in_decl_map(ndata->name, DECL_CLASS)) {
+				/* Ignore kernel classes */
+				if (!is_userspace_class(ndata->name, ndata->traits)) {
```
dburgener, line 335: Shouldn't this just be part of the `name_is_class()` else if clause two lines above so we fall through to the main else, rather than repeating that code here?  It seems more maintainable that way.

Either that or add it to the `system_r` exception at the top.

src/if_checks.c
```diff
@@ -400,7 +415,12 @@ struct check_result *check_name_required_but_not_used_in_if(const struct
 		return NULL;
 	}
 
-	struct name_list *names_to_check = get_names_in_node(node);
+	struct name_list *names_to_check;
+	if (dd->flavor == DECL_CLASS) {
```
dburgener, line 419: Why do we need special handling here instead of just teaching `get_names_in_node()` about classes?

src/startup.c
```diff
@@ -93,6 +94,75 @@ enum selint_error load_access_vectors_source(const char *av_path)
...
+		if (*pos != ' ' && *pos != '\t') {
+			ret = SELINT_PARSE_ERROR;
+			break;
+		}
+
+		while (*pos == ' ' || *pos == '\t') {
+			pos++;
+		}
+
+		if (!islower((unsigned char)*pos)) {
```
dburgener, line 131: Is it actually a requirement that permissions be all lower case?  I don't see that requirement documented in the notebook. I don't think I've ever seen an uppercase permission, and there may very well be a reason they can't exist, but I'm not sure exactly what it is, and I wouldn't want us to break on uppercase permissions if someone has a weird userspace object manager somewhere using them, assuming that's possible.

src/startup.c
```diff
@@ -93,6 +94,75 @@ enum selint_error load_access_vectors_source(const char *av_path)
...
+		while (isspace((unsigned char)*pos)) {
+			pos++;
+		}
+
+		if (*pos != '\0' && *pos != '#') {
+			ret = SELINT_PARSE_ERROR;
+			break;
+		}
+
+		if (*pos == '#' && NULL != strstr(pos, "userspace")) {
```
dburgener, line 153: So the heuristic here is that we're parsing the comment for "userspace" to know which ones are userspace classes?

That's a little iffy, but I suppose there may not be a better way.  Can we call that out in a comment somewhere prominent?  And probably document it in the README as well.   This seems like potentially confusing behavior for users if they happen to add a userspace class without the comment, so it should be clearly documented rather than buried in a parsing function.

src/maps.c
```diff
@@ -273,6 +274,56 @@ unsigned int decl_map_count(enum decl_flavor flavor)
...
+		userspace_class->val = 1;
+		HASH_ADD_KEYPTR(hh_userspace_class, userspace_class_map, userspace_class->key,
+				strlen(userspace_class->key), userspace_class);
+	} else {
+		userspace_class->val = 1;
+	}
+}
+
+no_sanitize_unsigned_integer_
+int is_userspace_class(const char *class_name, const struct string_list *permissions)
```
dburgener, line 296: I think this is how it works, but just to make sure I'm understanding correctly:

This function returns 1 if we've positively marked something as a userspace class, correct?  So if we encounter some class `foo` and don't know about it, we return false here, and then don't lint on it needing a requirement?

If so, I think that's the desired behavior.  I'd like to see that clarified in a comment.

I'm particularly concerned about the case where we didn't find the security_classes file for some reason.  I think in that case, we should warn (as the code does) and then skip the userspace checks.  It looks like everything is doing that, but I wanted to double check to make sure, and make sure that gets in a comment for posterity.  Maybe a comment by the warning would be helpful as well.

dburgener: Oh, actually, that's not the behavior.  Don't we hit the "Class %s is listed in require block but is not a userspace class" message if we fail to load the security classes?  I feel like if we didn't load any security classes, we should just skip checking the class requires, rather than reporting on every single class require.

I haven't thought this through thoroughly, but maybe just having this function always return 0 if there are no userspace classes would be sufficient?

dburgener (changes requested):

src/name_list.c
```diff
@@ -0,0 +1,222 @@
...
+	data->flavor = flavor;
+	data->traits = NULL;
+	struct name_list *ret = xmalloc(sizeof(struct name_list));
+	ret->data = data;
+	ret->next = NULL;
+
+	return ret;
+}
+
+struct name_list *name_list_from_decl(const struct declaration_data *decl)
```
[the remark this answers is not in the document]
cgzones: Since commit https://github.com/SELinuxProject/selint/commit/3f83758c1825c1e720867e18ab97ee18ef0eadf9 each declaration of the same statement in a require block gets parsed into its own node.

src/if_checks.c
cgzones: Merged the condition.

src/startup.c
cgzones: Changed to `isalnum(3)`.

src/startup.c
cgzones: Added a section to the ReadMe.

src/if_checks.c
cgzones: Turns out `get_names_in_node()` already supported that case; dropped the special handling here.

src/maps.c
cgzones: The function `is_userspace_class()` returns `1` if `class_name` is a known userspace class or it is the `system` class with a permission not to be known as a kernel permission, else `0`.
Added a global flag `userspace_class_support` whether `security_classes` was successfully parsed to skip checks depending on userspace class support, since there is an instance where a return value from `is_userspace_class()` of `0` skips a warning, and one for the return value of `1`.

src/startup.c
dburgener: This seems good, thanks.  I also added an [issue](https://github.com/SELinuxProject/selint/issues/279) for later follow-up to enhance this section and tell users how to get the best use out of SELint if they choose to violate those conventions.

PR #254, opened 2022-11-23
selint: warn on duplicate policy configuration files

cgzones:
Warn if the scanned directory contains multiple policy configuration files with the same name, in particular 'modules.conf', 'obj_perm_sets.spt' or 'access_vectors'.

Take the first found one.

Also avoid leaking filenames on subsequent encounters.


src/main.c
```diff
@@ -434,15 +434,30 @@ int main(int argc, char **argv)
 		} else if (source_flag
 		           && !strcmp(file->fts_name, "modules.conf")) {
 			// TODO: Make modules.conf name configurable
-			modules_conf_path = xstrdup(file->fts_path);
+			if (modules_conf_path) {
+				printf("%sWarning%s: multiple module.conf files found (using %s, ignoring %s)\n",
+				       color_warning(), color_reset(), modules_conf_path, file->fts_name);
```
dburgener, line 439: Shouldn't the last argument (for all three of these) be file->fts_path rather than fts_name?

I haven't tested this, but I imagine it will say "using path/to/modules.conf, ignoring modules.conf", which would be confusing.

dburgener (changes requested):

dburgener:
Merged, thanks!

PR #255, opened 2022-11-23
Reduce recursion in free_policy_node

cgzones:
Loop over the nodes in the next list instead of making a recursive call. Otherwise, together with the recursive call on the child nodes, the call stack might exceed the system limit, e.g. on s390x where the maximum size a process may lock into memory is quite small (8192 kB).

### DIFF
--- a/src/tree.c
+++ b/src/tree.c
@@ -369,12 +369,8 @@ struct policy_node *dfs_next(const struct policy_node *node)
 	}
 }
 
-enum selint_error free_policy_node(struct policy_node *to_free)
+static void free_single_policy_node_data(struct policy_node *to_free)
 {
-	if (to_free == NULL) {
-		return SELINT_BAD_ARG;
-	}
-
 	switch (to_free->flavor) {
 	case NODE_HEADER:
 		free_header_data(to_free->data.h_data);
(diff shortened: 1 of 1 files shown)


src/tree.c
```diff
@@ -369,12 +369,8 @@ struct policy_node *dfs_next(const struct policy_node *node)
 	}
 }
 
-enum selint_error free_policy_node(struct policy_node *to_free)
+static void free_single_policy_node(struct policy_node *to_free)
```
dburgener, line 372: Maybe "free_single_policy_node_data" would be a better name here?  Since it doesn't free the rest of the node.

src/tree.c
```diff
@@ -425,17 +421,25 @@ enum selint_error free_policy_node(struct policy_node *to_free)
...
+	}
 
-	free_policy_node(to_free->first_child);
-	to_free->first_child = NULL;
+	do {
+		struct policy_node *next = to_free->next;
 
-	free_policy_node(to_free->next);
-	to_free->next = NULL;
+		free_single_policy_node(to_free);
```
dburgener, line 435: The original order of recursing "next" first then "first_child" seems pretty weird.  I wonder if the original stack size problem would have even occurred if we'd recursed on "first_child" first.

dburgener:
This seems fine except for the one minor naming comment.  I do wonder if just switching the recursion order would work too, but this commit should be even better.

dburgener:
Merged, thanks!

PR #258, opened 2023-01-29
tests: clean generated testsuite artifacts

cgzones:
These three policy files are generated by the functional testsuite but not cleaned up.
Remove them with 'make clean', in particular to be able to build twice for Debian builds.


dburgener:
Merged.  Thanks!

PR #260, opened 2023-01-29
ci: force usage of DWARF4 by clang

cgzones:
The version of valgrind in the Ubuntu runners does not fully support DWARF5 yet.

Also bump the actions/checkout version to v3.


dburgener:
Looks great.  Merged.  Thanks!

PR #263, opened 2023-02-08
tests: really skip valgrind tests if valgrind is not available

cgzones:
I tried to be too clever with the one-liner, since running `command` does not work as expected when executed by bats (only directly in a shell).


dburgener:
This is clearer anyway.  Thanks!  Merged

PR #265, opened 2023-03-20
Support genfscon partial paths to be a quoted string

cgzones:
Required if the path contains a dot.


dburgener:
Looks great, thanks!

PR #272, opened 2023-10-05
Update GitHub action for releases

stevedlawrence:
Currently, the matrix configuration definitions and logic in the 'Release Info' step for which one of those configurations to use is far apart in the main.yml file. If the matrix configuration ever changes it can be difficult to also remember to update the 'Release Info' step, which can lead to failed releases.

To make this error less like, this adds a new 'is_release_job' matrix include that is set to false by default, but set to true for the one matrix configuration that should be used for building releases. Since these are near each other and all configured in the 'matrix' setting, it should hopefully be easier to remember to make updates if needed. Now matrix updates no longer require changes to the 'Release Info' step.

This also modifies the 'Release Info' step to remove the deprecated set-output command:

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,6 +24,10 @@ jobs:
       fail-fast: false
       matrix:
         cc: [ 'gcc', 'clang' ]
+        include:
+          - is_release_job: false
+          - is_release_job: true
+            cc: 'gcc'
     env:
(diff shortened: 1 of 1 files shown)


dburgener (approved):

dburgener:
Merged.  Thanks for the PR!

PR #273, opened 2023-12-06
Fix passing of argument with incompatible pointer type to fts_open()

JurajMarcin:
With GCC 14, the incompatible-pointer-types warning will turn into an error by default. Due to this, the compilation of selint with newer versions of GCC would fail.

This patch fixes 2 calls to `fts_open()` where this error occurs. As `fts_open()` expects an array of `char * const`, not `const char * const`, and the documentation does not guarantee that `fts_open()` will not modify the string, the argument needs to be copied (or initialized as an array instead of a pointer).


src/startup.c
```diff
@@ -39,7 +39,8 @@ enum selint_error load_access_vectors_kernel(const char *av_path)
 	}
 
 	enum selint_error r = SELINT_PARSE_ERROR;
-	const char *paths[2] = { av_path, NULL };
+	char *av_path_copy = xstrdup(av_path);
```
dburgener, line 42: Looks like this needs to be freed at the end of the function.

dburgener:
Seems reasonable except for the memory errors, thanks.

I'm curious though, since this is a warning upgrading to an error, I'm not getting the warning on the current code base.  I'm using gcc 13.2.1 and compiling with -Wall and -Wextra (per configure.ac).  The github tool chain also adds -WError.  So I don't understand why I'm not seeing the existing warning (probably as an error in CI).  Any ideas?

JurajMarcin:
In hindsight, the function calls no longer need to be wrapped in these macros.

dburgener:
Merged.  Thanks!

PR #276, opened 2023-12-29
Clang tidy cleanup

cgzones:


src/main.c
```diff
@@ -86,10 +86,12 @@ static void usage(void)
 
 }
 
-#define WARN_ON_INVALID_CHECK_ID(id, desc)\
```
dburgener, line 89: I the `do {} while(0)` trick seems tolerable when the macro isn't already a block, but I'm not really clear why it adds value here rather than just increasing the indent.  What exactly is clang-tidy upset about?

cgzones: It's not a clang-tidy lint, but while looking at the empty statements warnings I also looked for macros with control flow logic to protect them against potential dangling-else problems.

dburgener: Doesn't using the

```
if (condition) {
    code
}
```

style instead of 

```
if(condition)
    code
```

already protect against dangling else problems?

cgzones: Yes, it does.

dburgener: Then I think my preference would be to drop this commit.  I'd love to take the rest of the PR if you want to push a version dropping this commit.

dburgener:
Merged, thanks!

PR #277, opened 2023-12-29
man: create man page with asciidoctor instead of help2man

cgzones:
Support building man pages for cross builds.

Include more information, like 'Exit status' and 'Check IDs'.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,7 +48,7 @@ jobs:
     - name: install dependencies
       run: |
         sudo apt-get update -q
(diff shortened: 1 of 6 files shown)


dburgener:
I don't like that this duplicates the contents from the README.  I certainly am all for better man pages, but I'd prefer to have a single source of truth for documentation.  It's already bad enough that we duplicate the arguments in the code for `--help` and also in the README.  If we're going to migrate to a better documentation generator, I'd like it to reduce documentation duplication, and this seems to increase it.

PR #280, opened 2024-01-09
Use checked allocations

cgzones:


cgzones:
Missed during rebase of #250

dburgener (approved):

PR #282, opened 2024-02-23
Add check for fcontext lines containing only spaces

cgzones:
Currently file context definition lines containing only white spaces will result in an error:

    libraries.mod.fc:   130: (E): Bad file context format (E-002)

Support parsing those lines and issue a new style check.

Suggested in https://github.com/SELinuxProject/refpolicy/pull/763.


dburgener:
This looks great, thanks very much!

PR #283, opened 2024-02-26
Ignore fcontext lines containing only spaces

cgzones:
Currently file context definition lines containing only white spaces will result in an error:

    libraries.mod.fc:   130: (E): Bad file context format (E-002)

Ignore such lines.

Minimized alternative to #282 without a new check.

### DIFF
--- a/src/parse_fc.c
+++ b/src/parse_fc.c
@@ -14,6 +14,7 @@
 * limitations under the License.
 */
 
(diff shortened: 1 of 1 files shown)


dburgener:
This can be abandoned because #282 is merged, correct?

cgzones:
Yes.

PR #284, opened 2024-04-26
tests: drop trailing white spaces

cgzones:
Commit 6ff3d004 ("Add check for fcontext lines containing only spaces") added support for trailing white spaces in file context files together with some unit tests.  These unit tests contain such trailing white spaces, which break the current CI due to the white space check via `git diff-tree --check`.

Drop the trailing spaces (together with the corresponding test code) to please the CI, hoping regressions will not happen.


dburgener:
I'm not really a fan of making changes that actually hurt just to please broken CI.

What about #285 as an alternative?

PR #285, opened 2024-04-29
Run CI whitespace check only on changes introduced in latest PR

dburgener:


dburgener:
Another option is that git diff-tree takes a `path` argument, so we could exclude the offending files, but the syntax to do an exclusive bash glob is messy IMO.  This is more readable, and since everything comes in via PR anyways, I'm not sure we lose a lot by only checking the changes.

cgzones:
LGTM :+1:

PR #289, opened 2024-08-20
Allow M4 escaped strings inside QUOTED_STRING

bachradsusi:
When a file name in type transition rule used in an interface is same as a keyword, it needs to be M4 escaped so that the keyword is not expanded by M4, e.g.

-	filetrans_pattern($1, virt_var_run_t, virtinterfaced_var_run_t, dir, "interface")
+	filetrans_pattern($1, virt_var_run_t, virtinterfaced_var_run_t, dir, "``interface''")

Fixes:

    $ selint ./policy/modules/contrib/virt.if
    Note: Check E-007 is not performed because no permission macro has been parsed.
    ./policy/modules/contrib/virt.if:169: (F): syntax error, unexpected UNKNOWN_TOKEN (F-001)
      169 |     filetrans_pattern($1, virt_var_run_t, virtinterfaced_var_run_t, dir, "``interface''")
          |                                                                          ^
    ./policy/modules/contrib/virt.if:169: (F): Error: Invalid statement (F-001)
      169 |     filetrans_pattern($1, virt_var_run_t, virtinterfaced_var_run_t, dir, "``interface''")
          |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Error: Failed to parse files

PR #291, opened 2024-08-24
Support m4 quoted strings in interface call arguments

cgzones:
Required for escaped strings, that would otherwise be replaced by m4, e.g. "interface":

    filetrans_pattern(foo_t, bar_run_t, baz_run_t, dir, ``"interface"'')

Replaces: #289

Note: in contrast to #289 here the syntax is
```
``"interface"''
```
instead of
```
"``interface''"
```

### DIFF
--- a/src/parse.y
+++ b/src/parse.y
@@ -188,6 +188,7 @@
 %type<string> arg_list_item
 %type<sl> arg
 %type<sl> args
+%type<sl> arg_m4_quoted
 %type<string> mls_range
 %type<string> mls_level
 %type<string> mls_component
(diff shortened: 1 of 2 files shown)


dburgener:
Is there a reason we can't do both this and #289?  My inclination is to say that SELinux doesn't want to force either syntax on users, so it would be nice to handle both nesting options with quotes.  I haven't tried putting them together, but it looks like both PRs would be compatible with each other.

bachradsusi:
I like this approach. https://github.com/SELinuxProject/selint/pull/289 was a low effort which would allow to parse "``interface''" but it would parse also "``interface"  which is not correct. I've closed it.

I'm going to propose using ``"interface"'' in fedora-selinux/selinux-policy as suggested in this PR. It looks better for me.

I'm also working on sepolgen-ifgen patch which would be based on this idea - https://github.com/bachradsusi/SELinuxProject-selinux/commit/22b908e76f17851dac5f49dd4394068147c0b567

dburgener:
If the syntax here works for you, I think that's fine with me.  I don't want selint to be overly particular on matters of parsing - but that applies to what policy is used in practice rather than in theory.  This version is certainly nicer from an SELint perspective, and wrapping the m4 quotes outside seems to me to make more sense anyways.

I'll go ahead and merge this.  If anyone ever decides they actually want the syntax from #289 in practice, I'm happy to revisit that.  From an SELint parser perspective, it's nice to parse too much rather than too little, since parser errors are non-recoverable as SELint exists today.

PR #299, opened 2024-12-21
[RFC] Add support for gen_system_context

cgzones:
Support a wrapper for `gen_context(system_u:object_r:CTX, s0)` named `gen_system_context(CTX)`.

Support it as alternative to gen_context() in .te via the parser and in .fc files via extending the custom_fc_macros search logic.

I am not sure whether it's worth to have something non-reference style specific like this in SELint though.

Depends on #298.

### DIFF
--- a/src/lex.l
+++ b/src/lex.l
@@ -146,6 +146,7 @@ fs_use_task { return FS_USE_TASK; }
 define { return DEFINE; }
 gen_user { return GEN_USER; }
 gen_context { return GEN_CONTEXT; }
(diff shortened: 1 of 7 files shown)


dburgener:
What does use this?  Is this macro in some open-source policy?

Also, it looks like you didn't include the fixes to #298 on this branch, so it will need a rebase.

dburgener:
@cgzones ping on the above question

> What does use this? Is this macro in some open-source policy?

Just trying to understand the motivation for supporting this macro, since I've never heard of this particular macro.

PR #301, opened 2025-03-20
Add system permissions added to Linux for 6.15

cgzones:
For deciding whether a permission of the system class is a userspace or kernel one the kernel ones are hardcoded.  Add the ones to be introduced in Linux 6.15 with commit 2c2b1e059792 ("selinux: add permission checks for loading other kinds of kernel files").

See https://github.com/SELinuxProject/selinux-kernel/commit/2c2b1e059792f610bae4fee8ed517b8ce9c585fb


dburgener (approved):

dburgener:
Thanks!  I did the notebook update for these, but forgot about this special casing in SELint.  Good catch.

PR #302, opened 2025-04-09
tests: dump output of invoked commands

cgzones:
Dump the output of invoked commands in the functional tests, so that they are shown when the testcase fails, to ease debugging. (The output is not shown in case the testcase succeeds.)

Inspired due to a valgrind failure on armhf on Debian buildd.

### DIFF
--- a/tests/functional/end-to-end.bats
+++ b/tests/functional/end-to-end.bats
@@ -22,7 +22,7 @@ do_test() {
 	local EXPECT=$3
 	local ARGS=$4
 	run ${SELINT_PATH} -s -c tmp.conf ${ARGS} ./policies/check_triggers/${FILENAME} ./policies/check_triggers/modules.conf ./policies/check_triggers/obj_perm_sets.spt ./policies/check_triggers/access_vectors ./policies/check_triggers/security_classes
-	echo $output
+	echo ${output}
 	[ "$status" -eq 0 ]
 	count=$(echo ${output} | grep -o ${CHECK_ID} | wc -l)
 	echo "Status: $status, Count: $count (expected ${EXPECT})"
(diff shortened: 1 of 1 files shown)


tests/functional/end-to-end.bats
```diff
@@ -381,53 +386,62 @@ test_report_format_impl() {
...
 }
 
 @test "nesting_gen_req" {
 	run ${SELINT_PATH} -c configs/default.conf -e W-002 -E -s policies/misc/nesting.*
+	echo ${output}
 	[ "$status" -eq 0 ]
 	count=$(echo ${output} | grep -o "W-002" | wc -l)
 	echo "Status: $status, Count: $count (expected 1)"
-	echo $output
+	echo ${output}
```
dburgener: Shouldn't this and the remaining echo calls in this function just be dropped?  It looks like it's just printing the same output four times in a row.

tests/functional/end-to-end.bats
```diff
@@ -381,53 +386,62 @@ test_report_format_impl() {
...
 }
 
 @test "disable comment" {
 	run ${SELINT_PATH} -c configs/default.conf -F -e W-002 -E -s policies/misc/disable.*
+	echo ${output}
 	[ "$status" -eq 0 ]
 	count=$(echo ${output} | grep -o "W-002" | wc -l)
 	echo "Status: $status, Count: $count (expected 0)"
-	echo $output
+	echo ${output}
```
dburgener: As above, this one now seems redundant to the one above.

dburgener: (In this case, the subsequent ones in the function all follow a run, so for this test, it's just this line that I think needs dropping).

tests/functional/end-to-end.bats
cgzones: true, dropped superfluous statements

